PR #1, opened 2022-09-26
add check target to cmake that runs unit tests

majestrate:


tests/CMakeLists.txt
```diff
@@ -8,3 +8,5 @@ target_link_libraries(testAll PUBLIC
     session-util
     libsodium-internal
     Catch2::Catch2WithMain)
+
+add_custom_target(check COMMAND testAll)
```
jagerman, line 13: Are you sure this doesn't need to be `./testAll`?

majestrate: yes

majestrate: it will not build the binary if this target calls `./testAll`

PR #3, opened 2022-10-24
Config API

jagerman:


docs/config-merge-logic.md
```diff
@@ -0,0 +1,576 @@
...
+
+# Client update behaviour
+
+When there are multiple conflicting config messages clients resolve according to several rules.
+
+## Ignored updates
+
+Clients ignore updates according to two criteria:
+
+- "within five": a client ignores any update with a seqno value that is not within the last five
```
majestrate: note why it's 5. 

aka: this value controls conflict resolution depth via [insert metric here] so we chose 5 because ass number.

docs/config-merge-logic.md
```diff
@@ -0,0 +1,576 @@
...
+  seqnos.  For instance if current seqno is 123 then the client ignores any config message with a
+  seqno of 118 or earlier.
+
+- "ignore included": if a client sees two config messages with different seqno values and the
+  smaller one is contained within the diff of the larger one then the smaller one is stale and will
+  be ignored.
+
+## Simple updates
```
majestrate: stylistic nitpick: reword heading to be `non conflicting updates`

ghost:
Is semantic patch a useful concept for the config merge discussion?

There are some semantic patch implementations for different languages, like [Coccinelle for C ](https://coccinelle.gitlabpages.inria.fr/website/), or [JSON Patch](https://jsonpatch.com/). I see some overlaps here, hopefully, we can borrow some useful ideas or reuse some useful tools.

ghost:
Is there any plan to prevent a buggy client from propagating a malformed configuration to other clients? I'm guessing a local client might want to do a dry run and [validate the schema](https://json-schema.org/) before actually applying the update from a remote client?

majestrate:
we aren't using json for serialization as it is too permissive, not compact enough and the syntax is excessively complicated. we are using bittorrent encoding, see [BEP0003](https://www.bittorrent.org/beps/bep_0003.html#bencoding) bencode section.

majestrate:
additionally json patch doesnt provide enough conflict resolution for what we are looking to do. it does a single depth diff but we need a LOT more than that.

jagerman:
JSON isn't suitable for encoding in the first place because:
- binary data gets bloated by having to go through hex (2x size) or base64 (1.33x size) or base85 (1.25x size), and we expect to have a fair amount of binary data (keys, nonces, encrypted values, etc.) stored in config messages.
- given data doesn't have a unique encoding (e.g. `{"foo": 123}` and `{"foo":123}` are the same data, but different encodings; strings also have many possible equivalent encoded forms, e.g. `"foo\.bar"`, `"foo.bar"`, `"\u0066oo.bar"`, etc.).  Bencoding has exactly one valid representation of any given data structure, which is crucial here (because we rely on storage-server de-duping of identical stored data to not have to worry about clients racing to push the same content).

> I'm guessing a local client might want to do a dry run and [validate the schema](https://json-schema.org/) before actually applying the update from a remote client?

Yes: clients will ignore any incoming config message that doesn't conform to the requirements.  But we really want to be confident in the base layer that no such buggy content gets pushed (which is why all the Session clients will be using this library to produce messages) because if there was such a buggy version producing invalid data, that client would likely also be merging and propagating as non-buggy clients push data.  (This wouldn't be catastrophic in that it creates a runaway publishing loop, but would result in a perpetual split of the data that two different client versions would see).

docs/config-merge-logic.md
```diff
@@ -0,0 +1,609 @@
...
+
+To start with the inner config data (which will be under the `"&"` key of the outermost data
+message; see description in the [outer metadata](#config-message-outer-metadata), this is a dict
+which has byte string keys (of max length 128 bytes) and values which are any of:
+
+- integers, which are signed 64-bit quantities.
+- strings, which are of max length 4096 bytes.
+- sets of one or more values, which may contain unique integer or string values (but not other sets
+  or dicts).  These must be sorted by integer or byte value for ints or strings, respectively;
+  integers sort before strings if a list contains both.  Sets must contain at least one value: empty
```
ghost: If I understand correctly, you mean
> integers sort before strings if a **set** contains both

the draft says a **list** which is a bit confusing to me at first glance. I understand a `set` is encoded as a `list` after a second read though, maybe it's better to keep the wording consistent here?

docs/config-merge-logic.md
```diff
@@ -0,0 +1,609 @@
...
+                "changed": 1,
+                "foo": 123,
+                "nested": {
+                    "a": 1
+                }
+            },
+            "good": [99, 123, "Foo", "bar"],
+            "int1": 42,
+            "int2": 2,
+            "string2": "hello"
```
ghost: If I understand correctly, we are missing `"string3": "omg"` here?

docs/config-merge-logic.md
```diff
@@ -0,0 +1,609 @@
...
+2. All messages are sorted by their seqno (first) and hash (second).
+3. The "current data" is initialized from the full data of the highest value (i.e. highest hash of
+   the highest seqno).
+4. A replay set is constructed; this is ordered and must contain some reference to the message that
+   contains the diff (so that added/changed values can have their value retrieved from the
+   appropriate message).
+   - the "current" diff of every message is added to the replay sets (the seqno and hash are
+     extracted and computed from the message itself).
+   - past diffs in the messages (i.e. from the `"<"` list) are added to the replay set, *if* they
+     are not yet in the reply set, and have a seqno >= newseqno-5.  This should be done in order
```
ghost: typo: `reply set` should be `replay set`

same typo in
> Seqnos of that value are included for reply since all messages will have them, but are not included or needed for reply of future seqno messages.

docs/config-merge-logic.md
majestrate: > maybe it's better to keep the wording consistent here?

i think so. `set` would be consistent terminology.

fracting:
Science fiction: with fully homomorphic encryption, in theory, we can compare sequential numbers on the server side.
Not a serious proposal though, but it would be interesting to come back and re-evaluate the potential of FHE in a couple of years.

ghost:
There is an edge case I think the document doesn't explicitly cover:
What happens if a key is removed twice?

Suppose two clients push update with seqno=125 where:
* 125a sets `["int1"]` to `5`, and delete the `"string1"` key at the same time
* 125b sets `["int1"]` to `6`, and delete the `"string1"` key at the same time

By applying simple conflict resolution rule, `["int1"]` will be set to `6` after merging because 125b is higher than 125a.
However, the document doesn't say what will happen to `"string1"` in this case since there are two attempts to remove it.

My understanding is every operation (assign, deletion) should be idempotent, removing the same key twice should be equivalent to removing the same key once. It might seem too obvious so the document doesn't cover it, but it would be nice to have a unit test case on deleting the same key multiple times, so we can make sure the client won't crash when removing a non-exists key.

docs/config-merge-logic.md
jagerman: This is chosen by the caller now.

docs/config-merge-logic.md
jagerman: I spotted this and some other errors in here when using these as test cases, which are updated.

docs/config-merge-logic.md
jagerman: Thanks, fixed.

jagerman:
> There is an edge case I think the document doesn't explicitly cover: What happens if a key is removed twice?

Nothing; removing a key (or set value, for that matter) that doesn't exist does indeed just get ignored.  I've updated the doc to make this explicit.

.drone.jsonnet
```diff
@@ -252,4 +252,28 @@ local mac_builder(name,
...
+    steps: [
+      { name: 'submodules', commands: submodule_commands },
+      {
+        name: 'build',
+        environment: { SSH_KEY: { from_secret: 'SSH_KEY' } },
+        commands: [
+          'echo "Building on ${DRONE_STAGE_MACHINE}"',
+          // If you don't do this then the C compiler doesn't have an include path containing
+          // basic system headers.  WTF apple:
+          'export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)"',
```
majestrate, line 271: we should also put a precondition in `ios.sh` that makes it die if `$SDKROOT` isnt set.

jagerman: It's up to you to set up your compiler environment properly before compiling things (as is done here for drone). SDKROOT is one such way to do it for stock xcode clang, but if you want to use some other build mechanism then so be it.

jagerman:
Merging this now because this is the branch I'm using for other things anyway (i.e. this has effectively become `dev`).

PR #4, opened 2022-11-25
Add static builds for various platforms

jagerman:
Adds static bundles (that get uploaded via CI) for various platforms:
- Linux amd64/i386/arm64/armhf
- Windows x64/i686
- macOS (single multiarch lib for both x86_64 and arm64)
- Android (x86_64, armeabi-v7a, arm64-v8a, and x86)
- iOS (arm64 for devices + multiarch lib for x86_64 arm64 simulators)


majestrate:
i like the way things that dont belong in contrib arent in contrib and want to add that to lokinet.

utils/ios.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+set -e
+
+if ! [ -f LICENSE ] || ! [ -d include/session ]; then
+    echo "You need to run this as ./utils/ios.sh from the top-level libsession-util project directory"
```
majestrate: should this write to stderr as the one below does ?

utils/ios.sh
```diff
@@ -0,0 +1,79 @@
...
+set -x
+
+for i in "ios-arm64 OS64" "sim-arm64 SIMULATORARM64" "sim-x86_64 SIMULATOR64"; do
+    i=($i)
+    build="build-ios/${i[0]}"
+    platform="${i[1]}"
+    echo "Building ${i[0]} for $platform in $build"
+
+    ./utils/static-bundle.sh "$build" "" \
+        -DCMAKE_TOOLCHAIN_FILE=../../external/ios-cmake/ios.toolchain.cmake \
```
majestrate, line 39: use absolute path here as having it be a relative the way it is hard to understand.

utils/ios.sh
```diff
@@ -0,0 +1,79 @@
...
+
+
+# Create a '.xcframework' so XCode can deal with the different architectures
+xcodebuild -create-xcframework \
+    -library build-ios/ios-arm64/libsession-util.a \
+    -library build-ios/sim/libsession-util.a \
+    -output "$pkg_dir/libsession-util.xcframework"
+
+# Copy the headers over
+cp -rv include/session $pkg_dir/libsession-util.xcframework
```
majestrate: should those two paths to the xcframework be escaped the same?

utils/ios.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# The 'module.modulemap' is needed for XCode to be able to find the headers
+modmap="$pkg_dir/libsession-util.xcframework/module.modulemap"
```
majestrate, line 63: i feel like cmake should make this file honestly

utils/ios.sh
jagerman: Relative is easier when you are running the same commands manually while looking at the script (e.g. to debug something).

utils/ios.sh
jagerman: This is only something ios cares about so not worth the bother.

PR #5, opened 2022-12-04
Encryption & padding

jagerman:
Adds message encryption and prefix padding.


src/config/base.cpp
```diff
@@ -94,59 +190,159 @@ ConfigBase::ConfigBase(std::string_view dump) {
...
+    int ret = _keys_size;
+    _keys_size = 0;
+    return ret;
+}
+
+bool ConfigBase::remove_key(ustring_view key, size_t from) {
+    bool removed = false;
+
+    for (size_t i = from; i < _keys_size; i++) {
+        if (sodium_memcmp(_keys[0].data(), _keys[i].data(), KEY_SIZE) == 0) {
```
frtget, line 292: should this be `if (sodium_memcmp(key.data(), _keys[i].data(), KEY_SIZE) == 0) {` instead ?

jagerman: Yes indeed it should, good catch!

PR #11, opened 2023-01-19
Conversations

jagerman:
Adds Conversation config messages.

Still TODO:
- [x] finish the test code
- [x] add actual implementations of the C API functions
- [x] prune >30-day old entries

### DIFF
--- a/include/session/config/conversations.h
+++ b/include/session/config/conversations.h
@@ -7,13 +7,11 @@ extern "C" {
 #include "base.h"
 #include "profile_pic.h"
 
-
 enum CONVO_EXPIRATION_MODE {
     EXPIRATION_NONE = 0,
     EXPIRATION_AFTER_SEND = 1,
     EXPIRATION_AFTER_READ = 2,
 };
-
 
 typedef struct convo_one_to_one {
     char session_id[67];  // in hex; 66 hex chars + null terminator.
@@ -27,17 +25,17 @@ typedef struct convo_one_to_one {
 } convo_one_to_one;
 
 typedef struct convo_open_group {
-    const char* base_url; // null-terminated, always lower-case
-    const char* room; // null-terminated, always lower-case
-    const unsigned char* pubkey; // 32 bytes (not terminated, can contain nulls)
-    int64_t last_read; // ms since unix epoch
+    char base_url[321];        // null-terminated (max length 320), always lower-case
+    char room[151];            // null-terminated (max length 150), always lower-case
+    unsigned char pubkey[32];  // 32 bytes (not terminated, can contain nulls)
+    int64_t last_read;         // ms since unix epoch
 } convo_open_group;
 
 typedef struct convo_legacy_closed_group {
     char group_id[67];  // in hex; 66 hex chars + null terminator.  Looks just like a Session ID,
                         // though isn't really one.
 
-    int64_t last_read; // ms since unix epoch
+    int64_t last_read;  // ms since unix epoch
 } convo_legacy_closed_group;
 
 /// Constructs a conversations config object and sets a pointer to it in `conf`.
@@ -68,8 +66,8 @@ int conversations_init(
         char* error) __attribute__((warn_unused_result));
 
 /// Fills `convo` with the conversation info given a session ID (specified as a null-terminated hex
-/// string), if the conversation exists, and returns true.  If the conversation does not exist then `convo`
-/// is left unchanged and false is returned.
+/// string), if the conversation exists, and returns true.  If the conversation does not exist then
+/// `convo` is left unchanged and false is returned.
 bool convos_get_1to1(const config_object* conf, convo_one_to_one* convo, const char* session_id)
         __attribute__((warn_unused_result));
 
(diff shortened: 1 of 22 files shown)


include/session/config/conversations.h
```diff
@@ -0,0 +1,190 @@
...
+    int64_t last_read;
+
+    // expiration mode & time:
+    CONVO_EXPIRATION_MODE exp_mode;
+    int64_t exp_minutes;
+} convo_one_to_one;
+
+typedef struct convo_open_group {
+    const char* base_url; // null-terminated, always lower-case
+    const char* room; // null-terminated, always lower-case
```
mpretty-cyro: Is the `room` always lower case? (currently iOS forces the `base_url` to lowercase but not the `room`)

mpretty-cyro: Ah I saw the comment in `conversations.hpp` - feel free to just resolve this (I guess we will need to update the clients to force the `room` to lowercase? As whatever devices an open group wasn't joined on would have the value in lowercase regardless of what the casing was when initially joined)

include/session/config/conversations.h
```diff
@@ -0,0 +1,222 @@
...
+///             if (should_delete)
+///                 convos_iterator_erase(it);
+///             else
+///                 convos_iterator_advance(it);
+///         }
+///     }
+///     convos_iterator_free(it);
+///
+
+typedef struct convos_iterator {
```
majestrate: this iterator never is used as a non pointer so it can (and should) be a forward declared struct.

include/session/config/conversations.h
jagerman: I don't think this is our definitive list of which open groups you are in; that will likely live in a "groups" config message (and would be where the appropriate case is stored).  But it is case-insensitive on the SOGS side, so it seemed better to lower-case it here on input so that regardless of how a local client manages it, we'll have a consistent value here.

include/session/config/conversations.h
jagerman: Why?  Just to make it opaque?

majestrate:
On Fri, 20 Jan 2023 14:58:39 -0800
Jason Rhinelander ***@***.***> wrote:

> @jagerman commented on this pull request.
> 
> 
> 
> > +///
> +///     convo_one_to_one c1;
> +///     convos_iterator *it = convos_iterator_new(my_convos);
> +///     while (!convos_iterator_done(it)) {
> +///         if (convo_it_is_1to1(it, &c1)) {
> +///             bool should_delete = /* ... */;
> +///             if (should_delete)
> +///                 convos_iterator_erase(it);
> +///             else
> +///                 convos_iterator_advance(it);
> +///         }
> +///     }
> +///     convos_iterator_free(it);
> +///
> +
> +typedef struct convos_iterator {
> 
> Why?  Just to make it opaque?
> 

yes, with a C api a sign of a good design is that things that CAN be
opaque are always made opaque.

the only time i'd see a need to expose internals like that is when we
need to know the size of the struct or we have public members on the
struct we want callers to get at.

none of those are in play on this one so it's better to make it opaque.

-- 
~jeff

include/session/config/conversations.h
mpretty-cyro: I guess I figured adding a separate "groups" config would be redundant since we have this (I'm not sure what else we'd store in there 😕)

jagerman: I could see us storing some metadata in there such as the time that you joined, maybe a local alias for the group, and so on.

mpretty-cyro: Yea ok, I guess my worry is that any functionality we add for open group conversations we'd also want for non-open-group conversations as well and just end up with `CONVERSATIONS` and `CONVERSATIONS_2` config types

jagerman: My idea is that we have this one for tracking highly volatile conversation details (like last read timestamp), which we keep as small as possible, then put the details that don't change often into a "groups" config that contains the open group, closed group, legacy closed group fixed data.  (This is similar to what we have now with Contacts holding contact slow-changing data, and the last read timestamp going in here).

mpretty-cyro: Mmk, did you want to shift the disappearing messages setting into that type instead of having it in here then? (It might also make sense to call that one `CONVERSATIONS` and rename this one to something like `CONVOS_LAST_READ` then)

mpretty-cyro: (or are you thinking that the disappearing messages setting might change frequently enough as well?)

jagerman: Although thinking about it some more, perhaps rather than storing URL + room + pubkey we should just store a (smaller) hash of that in here to keep size reduced.

So you'd go pull your list of open groups from the Groups config, then would look up H(url | room | pubkey) in here.

jagerman: > Mmk, did you want to shift the disappearing messages setting into that type instead of having it in here then?

Yeah, actually, that's probably worthwhile.  I do want to keep this one (which gets re-push frequently) as small as possible.

mpretty-cyro: No worries, I'll let the guys know that the conversation stuff is still in flux in case they happen to start working on it

majestrate:
On Sun, 22 Jan 2023 13:58:41 -0800
Jason Rhinelander ***@***.***> wrote:

> @jagerman commented on this pull request.
> 
> 
> 
> > +
> +
> +typedef struct convo_one_to_one {
> +    char session_id[67];  // in hex; 66 hex chars + null terminator.
> +
> +    // milliseconds since unix epoch:
> +    int64_t last_read;
> +
> +    // expiration mode & time:
> +    CONVO_EXPIRATION_MODE exp_mode;
> +    int64_t exp_minutes;
> +} convo_one_to_one;
> +
> +typedef struct convo_open_group {
> +    const char* base_url; // null-terminated, always lower-case
> +    const char* room; // null-terminated, always lower-case
> 
> Although thinking about it some more, perhaps rather than storing URL
> + room + pubkey we should just store a (smaller) hash of that in here
> to keep size reduced.
> 
> So you'd go pull your list of open groups from the Groups config,
> then would look up H(url | room | pubkey) in here.
> 

this kind of optimizatgion has caused issues in the past (desktop
session). i think doing a (pubkey, room) tuple as the lookup key would
be better but at least please omit the url part of it.

-- 
~jeff

include/session/config/convo_info_volatile.hpp
```diff
@@ -0,0 +1,414 @@
...
+///
+/// 1 - dict of one-to-one conversations.  Each key is the Session ID of the contact (in hex).
+///     Values are dicts with keys:
+///     r - the unix timestamp (in integer milliseconds) of the last-read message.  Always
+///         included, but will be 0 if no messages are read.
+///     u - will be present and set to 1 if this conversation is specifically marked unread.
+///
+/// o - open group conversations.  This is a nested dict where the outer keys are the BASE_URL of
+///     the open group and the outer value is a dict containing:
+///     - `#` -- the 32-byte server pubkey
```
frtget, line 36: Would it make sense to swap the BASE_URL and the pubkey to prevent duplicates since a sogs can be reached via multiple BASE_URLs ?

frtget: I'm assuming changing the pubkey of a sogs would break things

majestrate: >I'm assuming changing the pubkey of a sogs would break things

yes, many things would. the stealth addresses ( the `15` prefixed ones ) of all the users would be the most notable of those changes.

include/session/config/convo_info_volatile.h
```diff
@@ -0,0 +1,219 @@
...
+///
+///     convo_info_volatile_1to1 c1;
+///     convo_info_volatile_iterator *it = convo_info_volatile_iterator_new(my_convos);
+///     while (!convo_info_volatile_iterator_done(it)) {
+///         if (convo_it_is_1to1(it, &c1)) {
+///             bool should_delete = /* ... */;
+///             if (should_delete)
+///                 convo_info_volatile_iterator_erase(it);
+///             else
+///                 convo_info_volatile_iterator_advance(it);
```
majestrate, line 171: this advancement of the iterator should be done outside the `if(convo_it_is_1to1(...` block as well.

include/session/config/convo_info_volatile.hpp
jagerman: > Would it make sense to swap the BASE_URL and the pubkey to prevent duplicates since a sogs can be reached via multiple BASE_URLs ?

No: You can use the same pubkey on multiple SOGS instances—and you might want to if, for instance, you decided to split up a single SOGS to multiple SOGS.  But it is impossible to use different pubkeys at the same URL.

include/session/config/convo_info_volatile.h
jagerman: Fixed in #15

PR #12, opened 2023-02-01
Add automatic compression/decompression

jagerman:
We use an embedded static version of zstd for compression, which gives us identical compression results for the same input (at least as long as we keep the same zstd version).

We always *try* to compress, but only actually use the compressed version if it ends up smaller (for very small messages, in particular, there often isn't enough repeated content to overcome the small overhead of compression).

### DIFF
--- a/src/config/base.cpp
+++ b/src/config/base.cpp
@@ -3,6 +3,7 @@
 #include <oxenc/hex.h>
 #include <sodium/core.h>
 #include <sodium/utils.h>
+#include <zstd.h>
 
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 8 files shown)


src/config/base.cpp
```diff
@@ -132,6 +169,25 @@ bool ConfigBase::needs_push() const {
     return !is_clean();
 }
 
+// Tries to compresses the message; if the compressed version (including the 'z' prefix tag) is
+// smaller than the source message then we modify `msg` to contain the 'z'-prefixed compressed
+// message and return true; otherwise we leave msg alone and return false.
```
frtget: you're not returning anything here

jagerman: Thanks!  An intermediate version had a bool return here but I removed it and forgot to update the comment!

Apparently I also forgot to add the test code.

PR #15, opened 2023-02-14
Group configs

### LINKED ISSUE #14: Version string exposed from the library
Just a function returning a string exposed from the library so we can print on the app startup which version of the library we are running (and include it in the logs if we ever run into an issue)

jagerman:
- Add new user groups config for tracking communities and legacy closed groups (and new closed groups in the future).
- Updated group terminology to match new Session terms: API bits named along the lines of "legacy closed group" are now "legacy group", and "open group" is now "community".
- Added new fields to contacts:
    - `hidden` (true = conversation hidden)
    - `priority` (higher = pinned higher; 0 = unpinned)
    - `exp_mode` expiration mode (named enum, defaulting to expiration_mode::none)
    - `exp_timer` the expiration timer, in minutes
- user groups similarly get `priority` for both communities/legacy groups, and legacy groups also get `hidden` and `disappearing_timer`.
- Abstracted the general "community" related settings (like base url, room, pubkey) into a new `community` class as it is shared between convo volatile and the new user groups configs.  (This also contains the general purpose static methods for community URL and name parsing/validation).
- profile picture `keylen` is removed; pic keys must now always be 32 bytes.
- the `pubkey` argument is gone from the `get_community` (previously `get_open`) method; we now will know the pubkey from the record if the record is found.
- `get_or_construct_community` still takes the pubkey, and the *given argument* pubkey becomes the pubkey in the object, even if what is stored is different.  (What is stored will be updated when the record is `set()`).
- fixed a bug where the outer service node entry (base URL key, plus "#" containing the pubkey) would remain in the config forever even after all rooms from that server were removed.
- exports the version (fixes #14)


include/session/config/contacts.h
```diff
@@ -19,6 +20,12 @@ typedef struct contacts_contact {
     bool approved;
     bool approved_me;
     bool blocked;
+    bool hidden;
+
+    int priority;
+
+    CONVO_EXPIRATION_MODE exp_mode;
```
mpretty-cyro, line 30: I get a build error here on iOS `Must use 'enum' tag to refer to type 'CONVO_EXPIRATION_MODE` - updating it to the below works though
```
enum CONVO_EXPIRATION_MODE exp_mode;
```

jagerman: Fixed (with a `typedef` in the expiring.h header so that the `enum` isn't needed here or elsewhere).

include/session/version.h
```diff
@@ -0,0 +1,19 @@
...
+
+/// libsession-util version triplet (major, minor, patch)
+extern const uint16_t LIBSESSION_UTIL_VERSION[3];
+
+/// Printable full libsession-util name and version string, such as `libsession-util v0.1.2-release`
+/// for a tagged release or `libsession-util v0.1.2-7f144eb5` for an untagged build.
+extern const char* LIBSESSION_UTIL_VERSION_FULL;
+
+/// Just the version component as a string, e.g. `v0.1.2-release`.
+extern const char* LIBSESSION_UTIL_VERSION_STR;
```
mpretty-cyro, line 15: When trying to access this from Swift I'm getting a "Undefined symbol" error (it's not strictly needed but I was going to include it in the logs in case it helps debugging down the line)

src/CMakeLists.txt
```diff
@@ -40,10 +42,44 @@ target_link_libraries(config
...
+            DEPENDS
+            "${CMAKE_CURRENT_SOURCE_DIR}/version.c.in"
+            "${PROJECT_SOURCE_DIR}/.git/index")
+    else()
+        message(STATUS "Git was not found or this is not a git checkout. Setting version tag to 'unknown'")
+        set(PROJECT_VERSION_TAG "nogit")
+        configure_file("${CMAKE_CURRENT_SOURCE_DIR}/version.c.in" "${CMAKE_CURRENT_BINARY_DIR}/version.c")
+    endif()
+endif()
+add_library(version version.c)
```
majestrate, line 79: `version.c` to `"${CMAKE_CURRENT_BINARY_DIR}/version.c"`

may fix morgan's link error issue.

jagerman: No, that wasn't it (the `version` lib was missing from the combined archive).

include/session/version.h
jagerman: Should be fixed now.

mpretty-cyro: Yep, can confirm this is fixed 👍

include/session/config/profile_pic.h
```diff
@@ -6,14 +6,16 @@ extern "C" {
...
-    // Null-terminated C string containing the uploaded URL of the pic.  Will be NULL if there is no
-    // profile pic.
-    const char* url;
-    // The profile pic decryption key, in bytes.  This is a byte buffer of length `keylen`, *not* a
-    // null-terminated C string.  Will be NULL if there is no profile pic.
-    const unsigned char* key;
-    size_t keylen;
+    // Null-terminated C string containing the uploaded URL of the pic.  Will be length 0 if there
+    // is no profile pic.
+    char url[224];
```
majestrate, line 15: i feel like the size of `url` and the value of `PROFILE_PIC_MAX_URL_LENGTH` will very easily get out of sync here. if that happens that feels like a very easy to footcannon regression.

majestrate: suggestion:

make `PROFILE_PIC_MAX_URL_SIZE` a macro (`_PROFILE_PIC_MAX_SIZE_VAL`) and have the size of `url` depend on it, `char url[1+_PROFILE_PIC_MAX_SIZE_VAL]`

then have `extern const size_t PROFILE_PIC_MAX_SIZE_URL` set to `_PROFILE_PIC_MAX_SIZE_VAL` in the compilation unit.

include/session/config/profile_pic.hpp
```diff
@@ -1,39 +1,57 @@
...
 namespace session::config {
 
-// Profile pic info.  Note that `url` is null terminated (though the null lies just beyond the end
-// of the string view: that is, it views into a full std::string).
+// Profile pic info.
 struct profile_pic {
-  private:
-    std::string url_;
-    ustring key_;
+    static constexpr size_t MAX_URL_LENGTH = 223;
```
majestrate, line 11: given changes from c header are applied, then this can be:

`static constexpr size_t MAX_URL_LENGTH = _PROFILE_PIC_MAX_URL_SIZE_VAL;`

include/session/config/community.hpp
```diff
@@ -0,0 +1,239 @@
...
+    // (https://example.com/r/SomeRoom?public_key=...).  The public_key is typically specified
+    // in hex (64 digits), but we also accept base64 (43 chars or 44 with padding) and base32z
+    // (52 chars) encodings (for slightly shorter URLs).
+    //
+    // The returned URL is normalized (lower-cased, and cleaned up).
+    //
+    // The returned room name is *not* normalized, that is, it preserve case.
+    //
+    // Throw std::invalid_argument if anything in the URL is unparseable or invalid.
+    static std::tuple<std::string, std::string, ustring> parse_full_url(std::string_view full_url);
```
mpretty-cyro, line 126: Can you please add a C interface for parsing a community URL? (figure it would be better to replace the custom Swift function with the libSession one)

tests/test_config_user_groups.cpp
```diff
@@ -0,0 +1,369 @@
...
+    auto c = groups.get_or_construct_legacy_group(definitely_real_id);
+
+    CHECK(c.session_id == definitely_real_id);
+    CHECK_FALSE(c.hidden);
+    CHECK(c.disappearing_timer == 0min);
+    CHECK(c.enc_pubkey.empty());
+    CHECK(c.enc_seckey.empty());
+    CHECK(c.priority == 0);
+    CHECK(c.name == "");
+    CHECK(c.members().empty());
```
mpretty-cyro, line 114: Is there a C interface for the user groups legacy group `members`?

tests/test_config_user_groups.cpp
```diff
@@ -0,0 +1,369 @@
...
+    CHECK(seqno == 1);
+
+    // Pretend we uploaded it
+    groups.confirm_pushed(seqno);
+    CHECK(groups.needs_dump());
+    CHECK_FALSE(groups.needs_push());
+
+    session::config::UserGroups g2{seed, groups.dump()};
+    CHECK_FALSE(groups.needs_push());
+    CHECK_FALSE(groups.needs_dump());
```
mpretty-cyro, line 199: Is this line redundant? (Have the same thing with a comment 2 lines down)

tests/test_config_user_groups.cpp
```diff
@@ -0,0 +1,369 @@
...
+                      });
+    }
+
+    auto x2 = g2.get_community("http://EXAMPLE.org:5678", "sudokuRoom");
+    REQUIRE(x2);
+    CHECK(x2->base_url() == "http://example.org:5678");
+    CHECK(x2->room() == "SudokuRoom");  // Case preserved from the stored value, not the input value
+    CHECK(x2->room_norm() == "sudokuroom");
+    CHECK(x2->pubkey_hex() == to_hex(open_group_pubkey));
+    CHECK(x2->priority == 14);
```
mpretty-cyro, line 256: @jagerman I seem to be getting a `priority` of `0` when retrieving the community via the C library here (I don't think I'm doing anything different...)

mpretty-cyro: The legacy group priority check seems to be working fine though (`CHECK(c1.priority == 3)`)

jagerman: Should be fixed in cefe8c2

include/session/config/profile_pic.h
jagerman: It cannot get out of sync because there's a static_assert that will fail to compile if it does.

include/session/config/community.hpp
jagerman: Added (in `session/config/community.h`)

jagerman: It also has one now to produce a URL from the parts.

tests/test_config_user_groups.cpp
jagerman: Will open a new PR with the additions.

Issue #13, opened 2023-02-12
Add property based testing using RapidCheck


ghost:
I'm investigating property-based testing for Oxen projects, I think libsession-util is a good start, and `session::bt::merge` is a good toy study case.

To use property-based testing, the first step is to define data generators, the second step is to define properties.

I implemented a toy data generator for a simplified version of `bt_value` type, next step is to integrate it into the rest of the code.

Once I have some working prototypes I'll create a draft PR for discussion.

In the meantime comments and feedback are appreciated!

ghost:
```
#include <rapidcheck.h>

#include <vector>
#include <algorithm>
#include <variant>

struct bt_value;

using bt_dict = std::map<std::string, bt_value>;
using bt_list = std::list<bt_value>;

using bt_variant = std::variant<
    std::string,
    std::string_view,
    int64_t,
    uint64_t,
    bt_list,
    bt_dict
>;

struct bt_value {
 bt_variant value;
};

bool operator==(const bt_value &lhs, const bt_value &rhs) {
  return std::visit([&](const auto &lhs_val, const auto &rhs_val) -> bool {
    using LHS = std::decay_t<decltype(lhs_val)>;
    using RHS = std::decay_t<decltype(rhs_val)>;
    if constexpr (std::is_same_v<LHS, std::string> &&
                  std::is_same_v<RHS, std::string>) {
      return lhs_val == rhs_val;
    } else if constexpr (std::is_same_v<LHS, std::string_view> &&
                         std::is_same_v<RHS, std::string_view>) {
      return lhs_val == rhs_val;
    } else if constexpr (std::is_same_v<LHS, int64_t> &&
                         std::is_same_v<RHS, int64_t>) {
      return lhs_val == rhs_val;
    } else if constexpr (std::is_same_v<LHS, uint64_t> &&
                         std::is_same_v<RHS, uint64_t>) {
      return lhs_val == rhs_val;
    } else if constexpr (std::is_same_v<LHS, bt_list> &&
                         std::is_same_v<RHS, bt_list>) {
      if (lhs_val.size() != rhs_val.size()) {
        return false;
      }
      auto lhs_iter = lhs_val.begin();
      auto rhs_iter = rhs_val.begin();
      for (; not (lhs_iter == lhs_val.end()); ++lhs_iter, ++rhs_iter) {
        if (not (*lhs_iter == *rhs_iter)) {
          return false;
        }
      }
      return true;
[...]

PR #17, opened 2023-02-16
[Draft] Initial integration of RapidCheck.

ghost:
I want to demonstrate the usage of RapidCheck in this draft PR and raise some discussions. See https://github.com/oxen-io/libsession-util/issues/13 for some context.

This PR checks properties like:

* self merge: `forAll x. merge(x, x) == x`. Edited: I think `self merge` is more appropriate than `idempotence` here. The combination of `self merge` and `associative` will eventually cover `idempotence`.
* associative: `forAll x. forAll y. forAll z. merge(merge(x, y), z) == merge(x, merge(y, z))`
* distributive: `forAll x. forAll y. forAll z. merge(x, merge(y, z)) == merge(merge(x, y), merge(x, z))`
* left priority of bt_dict: `forAll a. forAll b. if a and b share exactly the same keys, then merge(a, b) = a`

Edited: as well as more non-trivial properties. See patch content and CI for full details.

Done:
- [x] Basic test of `bt_dict operator==`
- [x] Good coverage of `session::bt::merge`.
- [x] Good coverage of `session::bt::merge_sorted`.

WIP (not published yet)
- [ ] Initial support of a configdata generator and tests.

### DIFF
--- a/tests/bt_cmp.cpp
+++ b/tests/bt_cmp.cpp
@@ -1,0 +1,55 @@
+#include <oxenc/bt_value.h>
+
+#include <stdexcept>
+
+using oxenc::bt_dict;
+using oxenc::bt_list;
+using oxenc::bt_value;
+
+bool bt_cmp(const bt_value& a, const bt_value& b) {
+    if (a.index() != b.index()) {
+        return a.index() < b.index();
+    }
+    switch (a.index()) {
+        case 0: {  // std::string
+            auto x = std::get<std::string>(a);
+            auto y = std::get<std::string>(b);
+            return x < y;
+        }
+        case 1: {  // std::string_view
+            auto x = std::get<std::string_view>(a);
+            auto y = std::get<std::string_view>(b);
+            return x < y;
+        }
+        case 2: {  // int64_t
+            auto x = std::get<int64_t>(a);
+            auto y = std::get<int64_t>(b);
+            return x < y;
+        }
+        case 3: {  // uint64_t
+            auto x = std::get<uint64_t>(a);
+            auto y = std::get<uint64_t>(b);
+            return x < y;
+        }
+        case 4: {  // bt_list
+            auto x = std::get<bt_list>(a);
+            auto y = std::get<bt_list>(b);
+            if (x.size() != y.size()) {
+                return x.size() < y.size();
+            }
+            return std::lexicographical_compare(x.begin(), x.end(), y.begin(), y.end(), bt_cmp);
+        }
+
+        case 5: {  // bt_dict
+            const auto& x = std::get<bt_dict>(a);
+            const auto& y = std::get<bt_dict>(b);
+
+            auto cmp_dict = [](const auto& a, const auto& b) {
+                return a.first < b.first || (a.first == b.first && bt_cmp(a.second, b.second));
+            };
+
+            return std::lexicographical_compare(x.begin(), x.end(), y.begin(), y.end(), cmp_dict);
+        }
+        default: throw std::runtime_error("Invalid bt_value type");
+    }
+}
--- a/tests/test_bt_merge.cpp
+++ b/tests/test_bt_merge.cpp
@@ -48,6 +48,59 @@ TEST_CASE("bt_dict merging", "[bt_dict][merge]") {
     CHECK(session::bt::merge(bt_dict{}, bt_dict{}) == bt_dict{});
 }
 
+namespace session::bt {
+TEST_CASE("bt_dict merging properties", "[bt_dict][merge]") {
+    rc::check("[bt_dict][merge] identity element", [](const bt_dict& x) {
+        auto e = bt_dict{};
+        REQUIRE(merge(e, e) == e);
+        REQUIRE(merge(x, e) == x);
+        REQUIRE(merge(e, x) == x);
+    });
+
+    rc::check(
+            "[bt_dict][merge] singleton dict",
+            [](const bt_dict& x, const std::string& y_k, const bt_value& y_v) {
+                auto y = bt_dict{{y_k, y_v}};
+                auto xy = merge(x, y);
+                auto yx = merge(y, x);
+
+                if (x.count(y_k) > 0) {
+                    REQUIRE(xy.size() == x.size());
+                    REQUIRE(yx.size() == x.size());
+
+                    REQUIRE(xy.at(y_k) == x.at(y_k));
+                    REQUIRE(yx.at(y_k) == y.at(y_k));
+                } else {
+                    REQUIRE(xy.size() == x.size() + 1);
+                    REQUIRE(yx.size() == x.size() + 1);
+
+                    REQUIRE(xy.at(y_k) == y.at(y_k));
+                    REQUIRE(yx.at(y_k) == y.at(y_k));
+                }
+            });
+
+    rc::check("[bt_dict][merge] self merge", [](const bt_dict& x) { REQUIRE(merge(x, x) == x); });
+
+    rc::check(
+            "[bt_dict][merge] associative",
+            [](const bt_dict& x, const bt_dict& y, const bt_dict& z) {
+                REQUIRE(merge(merge(x, y), z) == merge(x, merge(y, z)));
+            });
+
+    rc::check(
+            "[bt_dict][merge] distributive",
+            [](const bt_dict& x, const bt_dict& y, const bt_dict& z) {
+                REQUIRE(merge(x, merge(y, z)) == merge(merge(x, y), merge(x, z)));
+            });
+
+    rc::check("[bt_dict][merge] left priority", [](const bt_dict& x, const bt_dict& y) {
+        auto a = merge(x, y);
+        auto b = merge(y, x);
+        REQUIRE(merge(a, b) == a);
+    });
+}
+}  // namespace session::bt
+
 TEST_CASE("bt_list sorted merge", "[bt_list][merge]") {
     bt_list x{1, 2, 3, 5, 8, 13, 21};
     bt_list y{2, 4, 8, 16};
--- a/tests/test_bt_merge.cpp
+++ b/tests/test_bt_merge.cpp
@@ -1,11 +1,35 @@
+#include <rapidcheck.h>
+
 #include <catch2/catch_test_macros.hpp>
 
+#include "rc_gen_bt_value.hpp"
 #include "catch2_bt_format.hpp"
+#include "rapidcheck/catch.h"
 #include "session/bt_merge.hpp"
 
 using oxenc::bt_dict;
 using oxenc::bt_list;
 using oxenc::bt_value;
+
+TEST_CASE("bt_dict operator== properties", "[bt_dict][operator==]") {
+    rc::check(
+            "[bt_dict][operator==] order insensitive",
+            [](const std::string& x_k,
+               const std::string& y_k,
+               const bt_value& x_v,
+               const bt_value& y_v) {
+                auto x = bt_dict{{x_k, x_v}};
+                auto y = bt_dict{{y_k, y_v}};
+                auto xy = bt_dict{{x_k, x_v}, {y_k, y_v}};
+                auto yx = bt_dict{{y_k, y_v}, {x_k, x_v}};
+                if (x_k == y_k) {
+                    REQUIRE(xy == x);
+                    REQUIRE(yx == y);
+                } else {
+                    REQUIRE(xy == yx);
+                }
+            });
+}
 
 TEST_CASE("bt_dict merging", "[bt_dict][merge]") {
     bt_dict x{{"a", 1}, {"b", 2}, {"c", 3}};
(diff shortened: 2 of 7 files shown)


tests/test_bt_merge.cpp
```diff
@@ -1,12 +1,76 @@
...
+        return rc::gen::map(rc::gen::arbitrary<std::string>(), [](const std::string& s) {
+            return std::string_view(s);
+        });
+    }
+};
+
+Gen<bt_value> genBtValue() {
+    return gen::oneOf<bt_value>(
+            gen::map(
+                    gen::resize(20, gen::arbitrary<std::string>()),
```
majestrate: is there a way to give a random size instead of a hard coded `20`?

ghost: good point! yes, for std::string we can give a random size. for bt_dict and bt_list it's more tricky because the recursive structure quickly grows to a huge size and runs out of memory.

I'll implement a random-sized string generator.

ghost: is there a convenient way of printing a `bt_value`? I have a custom `print_bt_value()` function in https://github.com/oxen-io/libsession-util/issues/13 but I don't know if something similar is already implemented in libsessioin-util?

ghost: > good point! yes, for std::string we can give a random size. for bt_dict and bt_list it's more tricky because the recursive structure quickly grows to a huge size and runs out of memory.
> 
> I'll implement a random-sized string generator.

Updated. The way to generate a random-sized string is simple: just remove the gen::resize(20) wrapper.

majestrate: `bt_serialize` will give a std::string representation of it, but it will contain non printable characters if bytestrings have those in them. you could combine `bt_serialize` with a function to replace non printable chars with a placeholder char.

tests/test_bt_merge.cpp
```diff
@@ -1,12 +1,76 @@
...
+    }
+};
+
+Gen<bt_value> genBtValue() {
+    return gen::oneOf<bt_value>(
+            gen::map(
+                    gen::arbitrary<std::string>(),
+                    [](const std::string& s) { return bt_value{s}; }),
+            gen::map(
+                    gen::arbitrary<std::string_view>(),
```
majestrate: usage of `std::string_view` here is dubious as we have no concept of lifetime of the underlying data here.

tests/test_bt_merge.cpp
```diff
@@ -1,12 +1,76 @@
...
+                    [](const std::string_view& s) { return bt_value{s}; }),
+            gen::map(gen::arbitrary<int64_t>(), [](int64_t i) { return bt_value{i}; }),
+            gen::map(gen::arbitrary<uint64_t>(), [](uint64_t i) { return bt_value{i}; }),
+            gen::map(
+                    gen::resize(5, gen::container<bt_list>(gen::lazy(&genBtValue))),
+                    [](const bt_list& list) { return bt_value{list}; }),
+            gen::map(
+                    gen::resize(
+                            5,
```
majestrate: another hard coded length

ghost: Yea this is the tricky part.
> for bt_dict and bt_list it's more tricky because the recursive structure quickly grows to a huge size and runs out of memory.

I have to hardcode it other wise it will run out of memory. However, I can change it to a random number with an upper bound == 5.

majestrate: `random % upperbound`

ghost: I actually come up with a more interesting solution.
RapidCheck has a `gen::scale()` function which is aware of the size of the current generator, and scale it by a ratio before making another generator.

https://github.com/emil-e/rapidcheck/blob/master/doc/generators_ref.md#gent-scaledouble-scale-gent-gen

By using `gen::scale(0.5, )`, we are able to create complex structures which are both wide and deep, while the generator still guarantee to converge, terminate, without running out of memeory.

See https://gist.githubusercontent.com/fracting/1f3a9d12cbdc896b1e472167459e2801/raw/7b2c117f072fe4f77f42f03a28a1882099b032d9/bt_dict%2520test%2520case for one example generated by RapidCheck.

tests/test_bt_merge.cpp
ghost: Thanks!

tests/test_bt_merge.cpp
ghost: This one makes me think and I still don't have an answer, I lack the knowledge of some C++ fundamentals.

I think a safe way is like this:
```
template <>
class Arbitrary<std::string_view> {
  public:
    static rc::Gen<std::string_view> arbitrary() {
       return gen::just(std::string_view("string view placeholder"));
    }
};
``` 
which refers to some static string literals so we don't have to worry about the life cycle. Unfortunately, this can't generate random string_view dynamically.

I'm not sure the previous one is wrong either, not sure how RapidCheck manages the life cycle of std::string it creates, but it seems risky to rely on implementation details.

What's your suggestion? Any hints are appreciated.

ghost: Edit: this is wrong, ignore.

> ```
> rc::check("idempotence", [](const bt_dict& x) { REQUIRE(merge(x, x) == x); });
> ```
> 
> My understanding is, every time when a new `bt_dict x` is generated, a chain(or DAG) of generators manage their dependencies and the life cycle of the values they created. Any element in this chain should guarantee to be alive during a single test run in the lambda function `[](const bt_dict& x) { REQUIRE(merge(x, x) == x); }`, the RapidCheck framework should be able to take care of it. After a single test run is finished, everything is freed and a new `bt_dict x` is generated, a new test run is started, repeat the process until reaching 100 test runs. I think our usage of `std::string_view` should be safe in any single test run. Does that make sense?

ghost:
In theory, more test cases could be implemented, however the more complicated the test suite we have, the harder it is to maintain. I am going to stop at this point for `merge()` properties.

- [x] The next step is to implement some tests for `merge_sorted()` which could be more interesting.

Removed from TODO according to https://github.com/oxen-io/libsession-util/pull/17#issuecomment-1446356672
- [ ] After that I'll start investigating google/fuzztest and maybe create another draft PR.
(I personally feel google/fuzztest is less elegant than rapidcheck, but I haven't dive deep enough)

The overall purpose of this/these draft PR(s) is to:
1. demonstrate the potential of property-based testing
2. comparing different tools (RapidCheck, google/fuzztest, etc)
3. explore the proper scope of its usage in multiple Oxen projects

tests/test_bt_merge.cpp
```diff
@@ -24,14 +143,97 @@ TEST_CASE("bt_dict merging", "[bt_dict][merge]") {
...
+    });
+    return result;
+}
+
+bt_list gen_unique_sorted_bt_list() {
+    return map_to_bt_list(unique_sorted(*rc::gen::arbitrary<std::list<int64_t>>()));
+}
+
+// TODO: implement a generic version of compare()
+auto compare = [](const auto& a, const auto& b) {
```
majestrate: in C++20 there is `operator <=>`

majestrate:
i think we get the idea of what it can do.
my main worry is that this addition may remain unused unless people writing unit tests are excited to use it. unit tests tend to be a chore and making them easier to write is what matters.
i do not have a say in wrt that dimension at the moment.

ghost:
> i think we get the idea of what it can do. my main worry is that this addition may remain unused unless people writing unit tests are excited to use it. unit tests tend to be a chore and making them easier to write is what matters. i do not have a say in wrt that dimension at the moment.

Make sense, I share your concern as well. 

I think unit tests and property-based tests are not a replacement for each other.

When initiating a new feature with an uncertain design and rapid prototype, unit test is much more productive. When a module is semi-matured, property-based testing can ensure high reliability, and prevent accident breakage from future refactoring. When a subtle error is detected by property-based testing, the simplified test case can be added back to the regression suite. In other words, they are a good complementary of each other.

If any community contributor volunteers to add property-based testing for semi-matured modules, would that be welcome?
In that case, the core team does not have to change practice and slow down their development, while community contributors can add value in parallel without merging conflicts on actively developed code.

I think property-based testing can give us more confidence when the Oxen ecosystem scales up 10x to 100x, because at that scale more rare edge cases will occur to minority users. By investing in property-based testing in advance, we can be prepared for the potentially huge influx of users which might be unpredictable.

tests/test_bt_merge.cpp
ghost: > usage of std::string_view here is dubious as we have no concept of lifetime of the underlying data here.

@majestrate you are right. After adding more generic tests for merge_sorted(), I started to observe test failures caused by memory corruption, exactly like you said, `std::string_view` is the culprit. I replace the generator with a static one and no longer observe memory corruption. Unfortunately, I don't know how to generate a random `std::string_view` in an elegant way, so I'll leave it as a static placeholder for now.

tests/test_bt_merge.cpp
ghost: some submodule failed to build with C++20, I'll keep the old way for now. A generic version of compare and a generic version of bt_list merge_sorted() test is implemented.

ghost:
Update:

I started to use RapidCheck to test `ConfigMessage` and `MutableConfigMessage`.
Edit: prototype of configdata generator pushed.

The test case in https://github.com/oxen-io/libsession-util/pull/18 was the first failure discovered by RapidCheck.
I'll continue to investigate property based testing and continue update my findings.

The test itself was fairly simple:

```
TEST_CASE("config message property test", "[config][deserialization]") {
    rc::check("[config][deserialization] deserialize valid input", []() {
        bt_dict config_msg;
        auto old_seqno = *rc::gen::arbitrary<int64_t>();
        config_msg["#"] = bt_value{old_seqno};

        // my previous incorrect implementation of "valid data" triggered the segment fault bug
        // TODO: implement an invalid data generator to try to break the parser intentionally.
        config_msg["&"] = *rc::gen_valid_config_data();

        config_msg["<"] = bt_list();
        config_msg["="] = *rc::gen_valid_config_diff();

        auto dict_str = oxenc::bt_serialize(config_msg);
        RC_LOG(dict_str);
        MutableConfigMessage m{to_usv(dict_str)};
        auto m2 = m.increment();
        REQUIRE(m2.diff().empty());
    });
}
```


https://github.com/oxen-io/libsession-util/pull/17/commits/3895ae65fa06859fb0501dc9f3fa98e09015687a#diff-eaac86e9bacb5c65c34cf47f0d3492b9b9233ad0cb3fa13e55e8f6e00f2bb34dR33

ghost:
Not sure what's the right approach to fix compilation failure on g++, any advice is great appreciate.
Edit: fixed.

ghost:
Update:

Another potential case detected by rapid check: https://github.com/oxen-io/libsession-util/pull/23

tests/bt_cmp.cpp
```diff
@@ -0,0 +1,55 @@
...
+            const auto& y = std::get<bt_dict>(b);
+
+            auto cmp_dict = [](const auto& a, const auto& b) {
+                return a.first < b.first || (a.first == b.first && bt_cmp(a.second, b.second));
+            };
+
+            return std::lexicographical_compare(x.begin(), x.end(), y.begin(), y.end(), cmp_dict);
+        }
+        default: throw std::runtime_error("Invalid bt_value type");
+    }
```
majestrate, line 54: this can be redone using `std::visit` instead of calling `.index()` on `std::variant`.

we have a compat shim called `var::visit` in `oxenc/variant.h`

majestrate:
>comparing different tools (RapidCheck, google/fuzztest, etc)

i'd rather not pull in google code, we spent lots of time to clean our code bases of them 

i think this is fine to eventually merge since this caught a bug, i just want to make sure we can have these tests well maintained.

tests/bt_cmp.cpp
ghost: @majestrate thanks for the review! My previous version uses `std::visit`, but then some code failed to compile with old gcc. Then I tried to use `var::visit`, but `var::visit` accepts one argument only but I need two in `bt_cmp`. Any advice?

```
// Only handle a single variant here because multi-variant invocation is notably harder (and we
// don't need it).
template <class Visitor, class Variant>
constexpr auto visit(Visitor&& vis, Variant&& var) {
    return visit_helper(std::forward<Visitor>(vis), std::forward<Variant>(var),
            std::make_index_sequence<std::variant_size_v<std::remove_reference_t<Variant>>>{});
}
```

ghost:
> > comparing different tools (RapidCheck, google/fuzztest, etc)
> 
> i'd rather not pull in google code, we spent lots of time to clean our code bases of them
> 
> i think this is fine to eventually merge since this caught a bug, i just want to make sure we can have these tests well maintained.

Thanks! I'm not confident to make a formal pull request yet. There are some rapidcheck exceptions handling things I don't understand yet (potentially causing weird behaviors) and I haven't had time to learn more about how rapidcheck internal works.

PR #21, opened 2023-02-23
Obsolete hash tracking

### LINKED ISSUE #19: Track message hashes to delete
Change the `merge` API so that we take pairs of hash => data, then make any stale hashes available in some call after the merge so that Session clients can retrieve them and remove them from the server.

jagerman:
This adds tracking of obsolete hashes.  Doing so requires reworking some of the APIs around merging/pushing to now take the hash.

Fixes #19

### DIFF
--- a/include/session/config.hpp
+++ b/include/session/config.hpp
@@ -89,7 +89,11 @@ class ConfigMessage {
 
     bool verified_signature_ = false;
 
-    bool merged_ = false;
+    // This will be set during construction from configs based on the merge result:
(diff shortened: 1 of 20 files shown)


jagerman:
This now also adds the legacy group member api that was missing from #15 (added here because the changes here add test suite differences).

mpretty-cyro (changes requested):
Looks like iOS is getting a build error (`Unknown type name 'bool'`) - need to add `#include <stdbool.h>` to the top of `community.h`

PR #23, opened 2023-02-27
[DO NOT MERGE] Add config data test case - see comments for details

ghost:
I wrote some more property-based testing for config data (not published yet).

One of my tests discovered a "bug".

However, further investigation shows that my local repository was outdated, the "bug" was already fixed in

commit ca382d6fd0adc2e113098c7f0b303f0ecb259cf4
Author: Jason Rhinelander <jason@imaginary.ca>
Date:   Tue Feb 21 21:24:28 2023 -0400

    Fix segfault

    God knows how this was working at all.

I hope I can use this opportunity to demonstrate another case where a rare bug can be caught by property-based testing.

However, to be very honest, my investigation of property-based testing isn't as fruitful as I expected, it's time-consuming and I can't confidently argue that it is worth the dev team's time at this moment.

I still believe in its long-term value so I want to use this PR as a document for valuable cases.

My comment in the test case might answer Jason's git commit message `God knows how this was working at all.`
> // The actual value matters because it affects the order of hashes.

The bug fixed by Jason's commit can only be reproduced when lagged_diffs are ordered in a certain pattern. Different data value assignments result in different hashes and different orders of lagged_diffs. That's why traditional unit tests are not good at discovering such kinds of bugs and that's where property-based testing is powerful.

Feel free to discard or integrate the test case.

**do not merge** without dropping the commit with title `[DO NOT MERGE] ALERT ...`

Edited: Sorry if this is confusing:
In this draft PR I intentionally reverted ca382d6 to demonstrate a bug that's already fixed. It's a regression test conducted in the aftermath of ca382d6 to prevent similar bugs from coming back again in the future.

(I caught this bug because I forgot to sync my local repo and then rapidcheck found the bug.)

### DIFF
--- a/tests/test_configdata.cpp
+++ b/tests/test_configdata.cpp
@@ -1144,3 +1144,65 @@ TEST_CASE("config message example 4 - complex conflict resolution", "[config][ex
     CHECK(m_alt1.seqno() == 127);
     CHECK(m_alt1.hash() == m127.hash());
 }
+
+template <typename T, size_t N>
+const std::string_view array_to_sv(const std::array<T, N>& data) {
+    return to_sv(ustring_view{data.data(), data.size()});
+}
+
+inline bt_list lagged_from(MutableConfigMessage m) {
+    return {m.seqno(), array_to_sv(m.hash()), m.diff()};
+}
+
+TEST_CASE("merge test variant 1", "[config][merge]") {
+    MutableConfigMessage m0_x;
+    m0_x.data()[""] = "";
+    REQUIRE(m0_x.seqno() == 0);
+
+    MutableConfigMessage m1_xa{m0_x.serialize()};
+    REQUIRE(m1_xa.seqno() == 1);
+    m1_xa.data().erase("");
+
+    MutableConfigMessage m1_xb{m0_x.serialize()};
+    // The actual value matters because it affects the order of hashes.
+    m1_xb.data()[""] = 10;
+    m1_xb.data()["a"] = "y";
+
+    MutableConfigMessage m2_xab{{m1_xa.serialize(), m1_xb.serialize()}};
+    // There is a competition for the empty key "" between m1_xa and m1_xb, m1_xb wins.
+    // One way of verifing the winner is by checking the message data.
+    REQUIRE(m2_xab.seqno() == 2);
+    REQUIRE(printable(oxenc::bt_serialize(m2_xab.data())) ==
+            printable(oxenc::bt_serialize(m1_xb.data())));
+
+    // The other way of verifing the winner is by checking the order of lagged_diffs in m2_xab
+    bt_dict d2_xab;
+    bt_deserialize(to_sv(m2_xab.serialize()), d2_xab);
+    std::list<bt_list> lagged_2_xab{lagged_from(m0_x), lagged_from(m1_xa), lagged_from(m1_xb)};
+    REQUIRE(printable(oxenc::bt_serialize(lagged_2_xab)) ==
+            printable(oxenc::bt_serialize(d2_xab["<"])));
+
+    // another branch start from scratch
+    MutableConfigMessage m0_y;
+    // Occasionally equal to m1_xb.data();
+    m0_y.data()[""] = 10;
+    m0_y.data()["a"] = "y";
+
+    MutableConfigMessage m3{{m0_y.serialize(), m2_xab.serialize()}};
+    REQUIRE(m3.merged());
+    REQUIRE(m3.seqno() == 3);
+
+    bt_dict d3;
+    bt_deserialize(to_sv(m3.serialize()), d3);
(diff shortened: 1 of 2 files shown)


ghost:
This was the automated test that found the bug:

```
    rc::check("[config][merge] need a good name suggestion", []() {
        auto samples = *rc::gen_unique_valid_config_data_samples(3);

        MutableConfigMessage m0_x;
        m0_x.data() = pop_and_delete(samples);

        MutableConfigMessage m1_xa{m0_x.serialize()};
        MutableConfigMessage m1_xb{m0_x.serialize()};
        m1_xa.data() = pop_and_delete(samples);
        m1_xb.data() = pop_and_delete(samples);

        MutableConfigMessage m2_xab{{m1_xa.serialize(), m1_xb.serialize()}};

        MutableConfigMessage m0_y;
        m0_y.data() = m2_xab.data();

        MutableConfigMessage m3{{m2_xab.serialize(), m0_y.serialize()}};
        REQUIRE(m3.merged());
        REQUIRE(printable(bt_serialize(m3.data())) == printable(bt_serialize(m2_xab.data())));
    });
```

It automatically shrinks the test case to a fairly small one and eventually causes a segment fault.
Example output from rapidcheck:


```
!/Users/qhong/src/libsession-util-clean-up/tests/test_configdata_rc.cpp:225: FAILED:
  REQUIRE( printable(bt_serialize(m3.data())) == printable(bt_serialize(m2_xab.data())) )
with expansion:
  "d1:bl0:ee" == "d0:2:b(1:bl0:ee"

/Users/qhong/src/libsession-util-clean-up/tests/test_configdata_rc.cpp:225: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

In this PR I didn't pick the initializing value from rapid check directly, instead, I manually edited the initializing value to make the test case more human-readable. see the CI output:

```
/drone/src/tests/test_configdata.cpp:1206: FAILED:
  REQUIRE( printable(oxenc::bt_serialize(m3.data())) == printable(oxenc::bt_serialize(m2_xab.data())) )
with expansion:
  "d1:a1:ye" == "d0:i10e1:a1:ye"
```

tests/test_configdata.cpp
```diff
@@ -1144,3 +1144,71 @@ TEST_CASE("config message example 4 - complex conflict resolution", "[config][ex
...
+const std::string_view s_view(const std::array<T, N>& data) {
+    return to_sv(ustring_view{data.data(), data.size()});
+}
+
+inline bt_list lagged_from(MutableConfigMessage m) {
+    return {m.seqno(), s_view(m.hash()), m.diff()};
+}
+
+TEST_CASE("merge test variant 1", "[config][merge]") {
+    MutableConfigMessage m0_x;
```
majestrate, line 1158: i think if you value initialize `m0_x` here the unit test may pass. looks like it has a bunch of uninitialized memory in the failed test cases (maybe).

majestrate: rather, value initialize all your locals, it is something that is expected usually. i dont know if we assume that we need to value initialize  or not.

ghost: I think my way of initalizing `m0_x` is not different from other examples in the test suite like https://github.com/oxen-io/libsession-util/blob/dev/tests/test_configdata.cpp#L62

```
TEST_CASE("config pruning", "[config][prune]") {
    MutableConfigMessage m;
    m.data()["a"] = 123;
```

My test case addressed a bug **already fixed** in ca382d6fd0adc2e113098c7f0b303f0ecb259cf4

Sorry if I confused you, in this draft PR I intentionally reverted ca382d6fd0adc2e113098c7f0b303f0ecb259cf4 to demonstrate a bug that's already fixed. It's an "afterward" regression test to prevent similar bugs from coming back again in the future.

I caught this bug because I forgot to sync my local repo and then rapidcheck found the bug.

Hopefully I'm making it clearer for you, thanks!

@majestrate

ghost:
> ```
> template <typename T, size_t N>
> const std::string_view s_view(const std::array<T, N>& data) {
>     return to_sv(ustring_view{data.data(), data.size()});
> }
> 
> inline bt_list lagged_from(MutableConfigMessage m) {
>     return {m.seqno(), s_view(m.hash()), m.diff()};
> }
> ```
> 
> Does anyone mind suggesting a better name for `s_view`?

I call it `array_to_sv` now.

tests/test_configdata.cpp
jagerman: > i think if you value initialize `m0_x` here the unit test may pass. looks like it has a bunch of uninitialized memory in the failed test cases (maybe).

That won't matter here; the distinction between `T x{};` and `T x;` only exists when `T` has a trivial default constructor with trivial-constructible members (in other words: when it looks like a C struct).

PR #26, opened 2023-03-29
Add created & joined_at fields

jagerman:
These are both unix timestamps:
- Contacts `created` is when the contact record was created.
- legacy and community group `joined_at` is similar: when the group was "created", i.e. joined.


majestrate (approved):
reviewed, looks okay.

PR #28, opened 2023-04-13
Add better error diagnostics on failure to C API

jagerman:
The C++ API throws exceptions, but these were just being squelched into a single bool return; this changes most such places to also set `conf->last_error` to the error message when such an error occurs (and to clear `conf->last_error` when no error occurs).

### DIFF
--- a/src/config/contacts.cpp
+++ b/src/config/contacts.cpp
@@ -188,13 +188,16 @@ std::optional<contact_info> Contacts::get(std::string_view pubkey_hex) const {
 }
 
 LIBSESSION_C_API bool contacts_get(
-        const config_object* conf, contacts_contact* contact, const char* session_id) {
+        config_object* conf, contacts_contact* contact, const char* session_id) {
     try {
+        conf->last_error = nullptr;
         if (auto c = unbox<Contacts>(conf)->get(session_id)) {
             c->into(*contact);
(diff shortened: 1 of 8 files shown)


majestrate:
thought: 

what if we made a `struct error_info` that we can attach to conf instead? that way we can reuse it in other places too.

jagerman:
Seems like overkill considering the main point here is just for debugging/diagnostics.

src/config/contacts.cpp
```diff
@@ -188,13 +188,15 @@ std::optional<contact_info> Contacts::get(std::string_view pubkey_hex) const {
 }
 
 LIBSESSION_C_API bool contacts_get(
-        const config_object* conf, contacts_contact* contact, const char* session_id) {
+        config_object* conf, contacts_contact* contact, const char* session_id) {
     try {
         if (auto c = unbox<Contacts>(conf)->get(session_id)) {
```
frtget, line 194: forgot to clear conf->last_error here

majestrate: this is why i suggested a distinct type, that way we can use raii inside to prevent these kinds of foot cannons and others.

jagerman: Good catch!  Fixed.

PR #34, opened 2023-06-05
start of docs

darcys22:


darcys22:
Still working through this, thoughts are that there should be 2 api pages. One for the `C` headers and another for the `CPP` headers

darcys22:
Still left to do for C headers - `user_groups.h`

Should investigate a way to get structs into the docs aswell rather than just functions

docs/api/make-docs.sh
```diff
@@ -0,0 +1,58 @@
...
+else
+    echo "Can't determine your node_modules path; set NPM_PACKAGES or NODE_PATH appropriately" >&2
+    exit 1
+fi
+
+cp $npm_dir/docsify/node_modules/prismjs/components/prism-{json,python}.min.js api/vendor
+
+./api-to-markdown.py "$@"
+
+perl -ni.bak -e '
```
majestrate: may be nicer to use jinja2 templates to do this step, that would also let you do it all in that python script

majestrate: (I may volunteer myself this task if it doesn't conflict too hard)

PR #37, opened 2023-06-18
Update android ndk deb

darcys22:
Getting this error from the CI
```
Package google-android-ndk-installer is not available, but is referred to by another package.

This may mean that the package is missing, has been obsoleted, or

is only available from another source

However the following packages replace it:

  google-android-ndk-r25c-installer
```

android-ndk-r25c is the current long term release according to https://github.com/android/ndk/wiki and it appears that the source repos have depreciated the general package for specific versions. Going forward this will need to be kept up to date

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -316,7 +316,6 @@ local static_build(name,
   debian_pipeline(
     'Static Android',
     docker_base + 'android',
-    deps=['google-android-ndk-installer'],
     build=[
       'export JOBS=6',
       'export NDK=/usr/lib/android-ndk',


.drone.jsonnet
```diff
@@ -316,7 +316,7 @@ local static_build(name,
   debian_pipeline(
     'Static Android',
     docker_base + 'android',
-    deps=['google-android-ndk-installer'],
+    deps=['google-android-ndk-r25c-installer'],
```
jagerman: You ought to be able to just drop this line, and rely on whatever is in the `android` docker image (which currently is the r25c version).  That way we only have to update it there when it changes rather than multiple repos.

jagerman: (It's installed here: https://github.com/oxen-io/oxen-docker/blob/bf4571505aef62010e695daaff4239ee8905731c/ci/rebuild-docker-images.py#L321)

PR #38, opened 2023-06-18
update wine64 command

darcys22:
Appears that bookworm moved the binary for wine64
**https://packages.debian.org/bookworm/amd64/wine64/filelist**

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -152,7 +152,7 @@ local windows_cross_pipeline(name,
                    commands: [
                      apt_get_quiet + ' install -y --no-install-recommends wine64',
                      'cd build',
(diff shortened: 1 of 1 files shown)


.drone.jsonnet
```diff
@@ -152,7 +152,7 @@ local windows_cross_pipeline(name,
                    commands: [
                      apt_get_quiet + ' install -y --no-install-recommends wine64',
                      'cd build',
-                     'wine64-stable ./tests/testAll.exe --colour-mode ansi',
+                     '/usr/lib/wine/wine64 ./tests/testAll.exe --colour-mode ansi',
```
jagerman, line 155: The command you want is `wine-stable` (the 64 got dropped from the binary name).  I've added the change in #36 .

PR #42, opened 2023-08-03
New Groups support

jagerman:
Adds a group info config type, which tracks distributed info for v2 groups.

This requires introducing/using a few new concepts not currently used for user config messages:

- Multiple decryption keys.  User config doesn't do this at all (rather it generates a single decryption key from the private key for each namespace).

This doesn't yet add support for sharing and distributing those keys, just for having being able to load a config with a list of multiple possible keys.

- Config signing and verification.  For user configs this isn't done, since only the owner can actually encrypt/decrypt a config message, just being able to decrypt it is authentication enough.

This required various modifications to make the config library properly prevent modifications when we can't modify, and to properly follow that through in terms of merging, updates, etc.

# TODO:
- [x] storage server subaccount auth signature generation
- [x] storage server subaccount authentication helper
- [x] C API for user groups group_info
- [x] convo info volatile entry for group
- [x] support for wrapping/unwrapping a config in a protobuf wrapper, for existing config types (moved to PR #51 instead of here)
- [x] add members/keys size() functions for group types


include/session/config/groups/members.hpp
```diff
@@ -0,0 +1,356 @@
...
+
+    /// API: groups/Members::set
+    ///
+    /// Sets or updates the various values associated with a member with the given info.   The usual
+    /// use is to access the current info, change anything desired, then pass it back into set,
+    /// e.g.:
+    ///
+    /// ```cpp
+    ///     auto m = members.get_or_construct(pubkey);
+    ///     c.name = "Session User 42";
```
dr7ana, line 268: m.name?

include/session/config/groups/info.h
```diff
@@ -0,0 +1,269 @@
...
+LIBSESSION_EXPORT void contacts_set(config_object* conf, const contacts_contact* contact);
+
+// NB: wrappers for set_name, set_nickname, etc. C++ methods are deliberately omitted as they would
+// save very little in actual calling code.  The procedure for updating a single field without them
+// is simple enough; for example to update `approved` and leave everything else unchanged:
+//
+// contacts_contact c;
+// if (contacts_get_or_construct(conf, &c, some_session_id)) {
+//     const char* new_nickname = "Joe";
+//     c.approved = new_nickname;
```
frtget: mixed up approved and nickname

include/session/config/groups/info.hpp
```diff
@@ -0,0 +1,267 @@
...
+    ///
+    /// Returns the delete-before unix timestamp (seconds) for the group; clients should delete all
+    /// messages from the closed group with timestamps earlier than this value, if set.
+    ///
+    /// Returns std::nullopt if no delete-before timestamp is set.
+    ///
+    /// Inputs: none.
+    ///
+    /// Outputs:
+    /// - `int64_t` -- the unix timestamp for which all older messages shall be delete
```
frtget: messages -> attachments

include/session/config/groups/info.hpp
```diff
@@ -0,0 +1,267 @@
...
+    ///
+    /// Inputs:
+    /// - `timestamp` -- the new unix timestamp before which clients should delete attachments. Pass
+    /// 0
+    ///   (or negative) to disable the delete-attachment-before timestamp.
+    void set_delete_attach_before(int64_t timestamp);
+
+    /// API: groups/Info::get_delete_attach_before
+    ///
+    /// Returns the delete-before unix timestamp (seconds) for the group; clients should delete all
```
frtget: delete-before -> delete-attachments-before

include/session/config/groups/members.hpp
```diff
@@ -0,0 +1,356 @@
...
+    ///
+    /// Returns true if the user is already an admin *or* has a pending promotion to admin.
+    bool promoted() const { return admin || promotion_pending(); }
+
+    /// API: groups/member::info
+    ///
+    /// converts the member info into a c struct
+    ///
+    /// Inputs:
+    /// - `c` -- Return Parameter that will be filled with data in contact_info
```
frtget: contact_info -> member

include/session/config/groups/members.hpp
```diff
@@ -0,0 +1,356 @@
...
+    /// API: groups/Members::set
+    ///
+    /// Sets or updates the various values associated with a member with the given info.   The usual
+    /// use is to access the current info, change anything desired, then pass it back into set,
+    /// e.g.:
+    ///
+    /// ```cpp
+    ///     auto m = members.get_or_construct(pubkey);
+    ///     c.name = "Session User 42";
+    ///     members.set(c);
```
frtget: c should be m

src/config/groups/members.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/// Load _val from the current iterator position; if it is invalid, skip to the next key until we
+/// find one that is valid (or hit the end).
+void Members::iterator::_load_info() {
+    while (_it != _members->end()) {
+        if (_it->first.size() == 33) {
+            if (auto* info_dict = std::get_if<dict>(&_it->second)) {
+                _val = std::make_shared<member>(oxenc::to_hex(_it->first));
+                auto hex = oxenc::to_hex(_it->first);
```
frtget: the hex variable is not used

frtget:
should have done everything in a single review, sorry about that :)

src/config/groups/members.cpp
frtget: same in Contacts::iterator::_load_info()

src/config/groups/info.cpp
```diff
@@ -0,0 +1,107 @@
...
+void Info::set_expiry_timer(std::chrono::seconds expiration_timer) {
+    set_positive_int(data["E"], expiration_timer.count());
+}
+
+void Info::set_created(int64_t timestamp) {
+    set_positive_int(data["c"], timestamp);
+}
+
+std::optional<int64_t> Info::get_created() const {
+    if (auto* ts = data["c"].integer())
```
frtget, line 68: shouldn't we check whether this is >0 here and in get_delete_before and get_delete_attach_before ?

jagerman: It shouldn't be possible for it to be set to 0; the setter uses `set_positive_int` which removes the key entirely if it has a value of 0 (or negative, for that matter).  If some alternate/future version wants to explicitly set it as 0, well, that's on that future version of me (or whoever does it).

include/session/config/groups/info.h
jagerman: There's a whole bunch of wrong things in here that still need to be fixed; I probably shouldn't have committed the file at all in its current state.

dr7ana (approved):

src/config/groups/keys.cpp
```diff
@@ -0,0 +1,1594 @@
...
+    bool success = false;
+    bool compressed = false;
+    char type = static_cast<char>(ciphertext[0]);
+    ciphertext.remove_prefix(1);
+    switch (type) {
+        case 'X': compressed = true; [[fallthrough]];
+        case 'x': {
+            auto nonce = ciphertext.substr(0, crypto_aead_xchacha20poly1305_ietf_NPUBBYTES);
+            ciphertext.remove_prefix(crypto_aead_xchacha20poly1305_ietf_NPUBBYTES);
+            plain.resize(ciphertext.size() - OVERHEAD);
```
frtget, line 1215: should substract crypto_aead_xchacha20poly1305_ietf_ABYTES instead of OVERHEAD because the size of cyphertext was already shrunk by (1 + crypto_aead_xchacha20poly1305_ietf_NPUBBYTES)

jagerman: Good catch!  Fixed in #53

include/session/config/groups/keys.hpp
```diff
@@ -0,0 +1,651 @@
...
+    ///   for the group.  (This can be re-obtained from `pending_config()` if needed until it has
+    ///   been confirmed or superceded).  This data must be consumed or copied from the returned
+    ///   string_view immediately: it will not be valid past other calls on the Keys config object.
+    ustring_view rekey(Info& info, Members& members);
+
+    /// API: groups/Keys::key_supplement
+    ///
+    /// Generates a supplemental key message for one or more session IDs.  This is used to
+    /// distribute existing active keys to a new member so that that member can access existing
+    /// keys, configs, and messages.  Only admins can call this.
```
frtget, line 299: I'm not sure what you meant by "access[ing] existing keys" here. Removed in #64  because it seems redundant.

include/session/config/groups/keys.hpp
```diff
@@ -0,0 +1,651 @@
...
+            std::string_view session_id, bool write = true, bool del = false) const;
+
+    /// API: groups/Keys::swarm_verify_subaccount
+    ///
+    /// Verifies that a received subaccount signing value (allegedly produced by
+    /// swarm_make_subaccount) is a valid subaccount signing value for the given group pubkey,
+    /// including a proper signature by an admin of the group.  The signing value must have read
+    /// permission, but parameters can be given to also require write or delete permissions.  A
+    /// subaccount signing value should always be checked for validity using this before creating a
+    /// group that would depend on it.
```
frtget, line 359: "creating a group" as in adding a group to user configs ?

include/session/config/groups/keys.hpp
```diff
@@ -0,0 +1,651 @@
...
+    ///
+    /// There are two versions of this function: a static one callable without having a Keys
+    /// instance that takes the group id and user's session Ed25519 secret key as arguments; and a
+    /// member function that omits these first two arguments (using the ones from the Keys
+    /// instance).
+    ///
+    /// Inputs:
+    /// - `groupid` -- the group id/pubkey, in hex, beginning with "03".
+    /// - `session_ed25519_secretkey` -- the user's Session ID secret key.
```
frtget, line 368: Why don't we just take the session ID instead of the secret key ? We only use it to extract the pk and then remove the sign to get the xpk and use that.

src/config/base.cpp
```diff
@@ -455,6 +486,70 @@ void ConfigBase::load_key(ustring_view ed25519_secretkey) {
...
+    if (!_sign_sk)
+        throw std::runtime_error{"Cannot make a seed hash without a signing secret key"};
+    std::array<unsigned char, 32> out;
+    crypto_generichash_blake2b(
+            out.data(),
+            out.size(),
+            _sign_sk.data(),
+            32,  // Just the seed part of the value, not the last half (which is just the pubkey)
+            reinterpret_cast<const unsigned char*>(key.data()),
+            std::min<size_t>(key.size(), 64));
```
frtget, line 549: I think we should assert that key is smaller than 64 instead in order to avoid using the same key in different contexts by mistake by using several keys that start with the same 64 characters. I can add that to #64

src/config/groups/keys.cpp
```diff
@@ -0,0 +1,1594 @@
...
+    // compute:
+    token.resize(36);
+    std::memcpy(token.data(), sign_val.data(), 4);
+
+    // T = |S|, i.e. we have to clear the sign bit from our pubkey
+    std::array<unsigned char, 32> T;
+    crypto_sign_ed25519_sk_to_pk(T.data(), user_ed25519_sk.data());
+    bool neg = T[31] & 0x80;
+    T[31] &= 0x7f;
+    if (0 != crypto_scalarmult_ed25519_noclamp(to_unsigned(token.data() + 4), k.data(), T.data()))
```
frtget, line 676: Is there a reason why we don't want to clamp here ?
Tor and Lokinet both clamp blinding factors but I'm not sure whether it really makes the DH key exchange more secure.
I don't think it protects against small subgroup attacks because `a` is a multiple of the cofactor so `±akB` should either be the identity element or be in the main subgroup anyway.
Do you think we should clamp just in case we decide to reuse this blinded keypair for X25519 key exchange in the future for some reason ?

include/session/config/groups/keys.hpp
```diff
@@ -0,0 +1,651 @@
...
+    ///   messages.
+    ///
+    /// This method is closely coupled to the group's Info and Members configs: it updates their
+    /// encryption keys and sets them as dirty, requiring a re-push to re-encrypt each of them.
+    /// Typically a rekey is performed as follows:
+    ///
+    /// - `rekey()` is called, returning the new keys config.
+    /// - `info.push()` is called to get the new info config (re-encrypted with the new key)
+    /// - `members.push()` is called to get the new members config (using the new key)
+    /// - all three new configs are pushed (ideally all at once, in a single batch request).
```
frtget, line 280: Shouldn't this be a sequence request instead of a batch request ?

src/config/groups/keys.cpp
frtget: It seems the person who implemented the key blinding in tor wasn't sure either.
@jagerman I see you're the one who added the code that clamps the blinding factor in lokinet so I'm curious to know the reason why it's done if you know it.

jagerman: The short answer is that it "doesn't work" with clamping half the time, because when we have to use `-a` instead of `a`, clamping it changes the value and breaks the `-A = -aG` relationship that we're depending on.  There's a fair bit to unpack behind this though, so let me elaborate.

I wish we didn't need that relationship at all: but unfortunately it's an artifact of a suboptimal choice (which unfortunately followed Signal) of having Session IDs be the X25519 pubkey conversions of Ed25519, rather than direct Ed25519 keys.  Unfortunately that Ed->X conversion loses the sign bit from the Ed25519 pubkey, which is what causes all this trouble: when reversing the Ed->X conversion we get ourselves to Ed=±sqrt(...) and there's no way to know which one was the correct Ed pubkey.  So what we're doing here is effectively always signing *as if* the real pubkey that yielded the Session ID was in fact always the positive of the two possible Ed25519 pubkeys, when in reality half the time it is negative.

In order to do that when we actually have a negative pubkey, however, we have to effectively produce a signature from `-a` rather than from `a` -- but that won't work if we clamp `-a`.

To illustrate:

```Python
#!/usr/bin/python3

from nacl.signing import SigningKey
import nacl.bindings as sodium

for seed31 in (b'\0', b'\4'):
    seed = SigningKey(b"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0" + seed31)
    pubkey = seed.verify_key

    # This gives the private scalar, i.e. "a" for keypair a/A that yields aG = A.  (That's not its primary usage in
    # libsodium, and thus the function name here is understandably confusing).
    a = seed.to_curve25519_private_key().encode()

    print(f"\nPubkey:     {pubkey.encode().hex()}")
    Ac = sodium.crypto_scalarmult_ed25519_base(a)
    An = sodium.crypto_scalarmult_ed25519_base_noclamp(a)
    print(f"A noclamp:  {An.hex()}")
    print(f"A clamped:  {Ac.hex()}")

    if Ac[31] & 0x80:
        a_neg = sodium.crypto_core_ed25519_scalar_negate(a)
        Ac_neg = sodium.crypto_scalarmult_ed25519_base(a_neg)
        An_neg = sodium.crypto_scalarmult_ed25519_base_noclamp(a_neg)
        print(f"-A noclamp: {An_neg.hex()}")
        print(f"-A clamped: {Ac_neg.hex()}")
```
produces:
```
Pubkey:     3b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29
A noclamp:  3b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29
A clamped:  3b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29

Pubkey:     fd50b8e3b144ea244fbf7737f550bc8dd0c2650bbc1aada833ca17ff8dbf329b
A noclamp:  fd50b8e3b144ea244fbf7737f550bc8dd0c2650bbc1aada833ca17ff8dbf329b
A clamped:  fd50b8e3b144ea244fbf7737f550bc8dd0c2650bbc1aada833ca17ff8dbf329b
-A noclamp: fd50b8e3b144ea244fbf7737f550bc8dd0c2650bbc1aada833ca17ff8dbf321b
-A clamped: d1cd8406d8f93617b10a91a20c4126c08bb4f0fb23807fcfd25e939b7a9924ea
```
(Keep in mind these are encoded in little-endian, and so the sign bit is the most significant bit of the *last* byte, i.e. the second-last hex digit).

So in short, without clamping we can keep up the illusion that we were actually the positive alternative of the two pubkeys that could have given the X25519, even when we aren't.

jagerman: > Do you think we should clamp just in case we decide to reuse this blinded keypair for X25519 key exchange in the future for some reason ?

I really wish we could, because you're right, we're okay for Ed25519 signatures, but definitely don't want to go back to X again with this derived key.

frtget: We are talking about different things. I was talking about clamping the blinding factor `k`, not clamping `t`.

After some more digging it turns out that the security provided by clamping scalars is not preserved by multiplication operations on scalars because they are done modulo `l` (the order of the main subgroup).
For a point `P` with a torsion component (i.e. not in the main subgroup):
`(-a mod l).P ≠ -(a.P)`
and `(k*t mod l).P ≠ k.(t.P)`
Which means that even though `a` was already clamped during key generation, `-a mod l` and `±ak mod l` as private key scalars are vulnerable to small subgroup attacks and **clamping `k` is useless**.

I think the reason why it was done in tor is that the developers weren't sure about the impact of clamping on the security of their key blinding scheme at the time and it seems the reason why it was added to lokinet was to make the crypto identical to tor's.

My understanding is that if we really wanted to make sure that this derived scalar is safe against such attacks we would have to use its [torsion safe representative](https://moderncrypto.org/mail-archive/curves/2017/000866.html) (which is an element of GF(`8l`)) or do arithmetic operations modulo the order of the entire elliptic curve group `8l` (which is effectively the same thing when `a` is a multiple of the cofactor) so that:
`kt.P = ±k.(a.P)`
`P` can be written as `Q + T` where `Q` is a point in the main subgroup and `T` is a point of small order.
`±k.(a.(Q + T)) = ±k.(a.Q + a.T) = ±k.(a.Q + ∞) = ±k.(a.Q)` because `a` is a multiple of the order of `T`.
The torsion component is absorbed and the resulting point does not leak any information about our private scalars.

But that's not supported by libsodium and, in my opinion, not necessary because if we ever decide to reuse this keypair for DH we would have to use either the "noclamp" scalar multiplication function which fails when given a point that is not in the main subgroup or the ristretto functions which, by design, can't decode input bytes to a point outside of the main subgroup.
(Everything else clamps the input scalar and breaks the `kt.B = b.(k.T)` relationship we need where b/B is the other party's keypair.)
So, as far as I know, it would still be safe to use this derived scalar as is without caring about any of the above. Unless we switch to another crypto library.

In conclusion, clamping `k` is useless and there is not much we can do for now except maybe adding some comments to make it clear that these derived keys should never be reused for DH without extra precautions because they are vulnerable to small subgroup attacks.

venezuela01:
We need more experts to review the design and implementation. I hope @KeeJef will publish more internal design drafts before an actual software release. We have seen examples in the past where in-house developed code was vulnerable to algebraic attacks due to subtle mistakes by the core team. This situation further evidences the need for more external volunteers like @frtget to provide valuable opinions.

PR #51, opened 2023-09-06
Protobuf handling

dr7ana:


jagerman:
~~TODO: we need to squash these changes together (had to update the protobufs again, for the downgrade).~~

jagerman:
Now squashed and rebased on top of #59 (which will most likely be merged before this PR).

PR #53, opened 2023-09-08
Fix plaintext size calculation

jagerman:
ciphertext has already been reduced by 1 +
crypto_aead_xchacha20poly1305_ietf_NPUBBYTES at this point, but that wasn't being accounted for in the size calculations.

(Thanks to @ftrget for finding this one!)


venezuela01:
typo, should be @frtget instead.

PR #56, opened 2023-09-18
feat: add make_dump() to ConfigBase and Keys classes

Bilb:


jagerman:
Rename it to `make_dump` instead of `debug_dump` as it gets used internally for `dump()` itself.  Also made it `const`.

jagerman (approved):

PR #59, opened 2023-09-21
Add signing and verification to encrypt_message/decrypt_message

jagerman:
This redoes how we encode the result of encrypt_message to add sender identify and verification (entirely contained within libsession).  These functions now produce/consume:

    encrypted(
        padded(
            bt-encoded(
                possibly-compressed-data,
                sender-ed25519-pubkey,
                ed25519-signature
            )
        )
    )

with signature verification on decryption.

The decryption function thus now returns a pair instead of an optional value: the session id, and the original value, assuming that value was signed by the session id.  Upon failure for any reason, an exception is thrown (rather than, previously, returning a nullopt without any way to get an informative reason for the failure for diagnostics).

### DIFF
--- a/include/session/config/groups/keys.hpp
+++ b/include/session/config/groups/keys.hpp
@@ -237,7 +237,8 @@ class Keys final : public ConfigSig {
     /// if there are no encryption keys at all.  (This is essentially the same as `group_keys()[0]`,
     /// except for the throwing and avoiding needing to constructor a vector).
     ///
-    /// You normally don't need to call this; you can just use encrypt_message() instead.
+    /// You normally don't need to call this; the key is used automatically by methods such as
+    /// encrypt_message() that need it.
     ///
     /// Inputs: none.
     ///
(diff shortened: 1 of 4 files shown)


include/session/config/groups/keys.hpp
```diff
@@ -606,58 +607,88 @@ class Keys final : public ConfigSig {
...
+    ///
+    /// This method is passed a binary value containing a group message (typically a serialized
+    /// protobuf, but this method doesn't care about the specific data).  That data will be, in
+    /// order:
+    /// - compressed (but only if this actually reduces the data size)
+    /// - signed by the user's underlying session Ed25519 pubkey
+    /// - tagged with the user's underlying session Ed25519 pubkey (from which the session id can be
+    ///   computed).
+    /// - all of the above encoded into a bt-encoded dict
+    /// - suffix-padded with null bytes so that the final output value will be a multiple of 256
```
frtget, line 620: are you using different padding methods for [config messages](https://github.com/oxen-io/libsession-util/blob/f36a8df21ef9c2908e7bd390b96baa2900666517/include/session/config/encrypt.hpp#L84-L108) and regular messages on purpose ?

PR #62, opened 2023-10-03
fix: remove set_name from the members struct

Bilb:
the name set has to be made through the field and members.set


jagerman:
Fixed via cda4d32

PR #69, opened 2023-10-20
Fixed an issue where the auth_data wasn't getting cleared correctly

mpretty-cyro:
When calling `ugroups_group_set_kicked()` followed by `user_groups_set_group()` in the C API the `auth_data` wasn't getting cleared.

### DIFF
--- a/src/config/user_groups.cpp
+++ b/src/config/user_groups.cpp
@@ -432,6 +432,8 @@ void UserGroups::set(const group_info& g) {
         info["K"] = ustring_view{};
         if (g.auth_data.size() == 100)
             info["s"] = g.auth_data;
+        else
+            info["s"] = ustring_view{};
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/config/user_groups.cpp
```diff
@@ -432,6 +432,8 @@ void UserGroups::set(const group_info& g) {
         info["K"] = ustring_view{};
         if (g.auth_data.size() == 100)
             info["s"] = g.auth_data;
+        else
+            info["s"] = ustring_view{};
```
jagerman: This will set an empty string in the serialized config, but the value will still be there; we should just erase the key completely with:
```suggestion
            info["s"].erase();
```

jagerman (approved):

src/config/user_groups.cpp
mpretty-cyro: Ah no worries - will try to remember in the future 😂

PR #72, opened 2023-11-03
Added the `removed_status` value and a way to retrieve the current `GROUP_KEYS` generation

mpretty-cyro:
Added a new removed value to enable admins to better recover from failing to properly remove members from a group, the value indicates whether just the member should be removed or if their messages should also be removed alongside them

Also added a `current_generation` function which retrieves the current `GROUP_KEYS` generation value

### DIFF
--- a/include/session/config/groups/keys.h
+++ b/include/session/config/groups/keys.h
@@ -301,8 +301,7 @@ LIBSESSION_EXPORT bool groups_keys_key_supplement(
 ///
 /// Oututs:
 /// - `int` -- latest keys generation number on success; returns `-1` on failure.
-LIBSESSION_EXPORT int groups_keys_current_generation(
-        config_group_keys* conf);
+LIBSESSION_EXPORT int groups_keys_current_generation(config_group_keys* conf);
 
 /// API: groups/groups_keys_swarm_make_subaccount
 ///
--- a/include/session/config/groups/keys.h
+++ b/include/session/config/groups/keys.h
@@ -292,6 +292,18 @@ LIBSESSION_EXPORT bool groups_keys_key_supplement(
         unsigned char** message,
         size_t* message_len);
 
+/// API: groups/groups_keys_current_generation
+///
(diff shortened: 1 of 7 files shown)


include/session/config/groups/keys.h
```diff
@@ -292,6 +292,17 @@ LIBSESSION_EXPORT bool groups_keys_key_supplement(
...
 
+/// API: groups/groups_keys_current_generation
+///
+/// Returns the current generation number for the latest keys message.
+///
+/// Inputs:
+/// - `conf` -- [in] Pointer to the config object
+///
+/// Oututs:
+/// - `int` -- latest keys generation number on success; returns `-1` on failure.
```
jagerman: I don't think it's possible for this to actually error (and thus return -1), given the implementation.

src/config/groups/keys.cpp
```diff
@@ -1574,6 +1574,15 @@ LIBSESSION_C_API bool groups_keys_key_supplement(
     }
 }
 
+LIBSESSION_EXPORT int groups_keys_current_generation(config_group_keys* conf) {
+    try {
+        return unbox(conf).current_generation();
+    } catch (const std::exception& e) {
+        set_error(conf, e.what());
+        return -1;
+    }
```
jagerman:
```suggestion
    return unbox(conf).current_generation();
```

(Continuing the other comment) ... there's nothing in here that can actually throw (it can crash if given an invalid pointer, but that's not something an exception can catch).

include/session/config/groups/keys.h
jagerman: (I pushed a fix)

src/config/groups/keys.cpp
jagerman: (pushed)

mpretty-cyro: Good to know - I'd just duplicated and modified what the other functions were doing 😂

PR #74, opened 2023-11-17
Add Multi-encrypt functions

jagerman:
This allows efficiently encrypting 1-to-N or N-to-N messages in a single message, such as is used in the group keys message.

This adds both a general C++ interface (which is now used by the group keys code) for constructing such encrypted messages generically, and adds a version suffixed with `_simple` that produces ready-to-go bt-encoded binary values containing the nonce + encrypted values.

### DIFF
--- a/include/session/multi_encrypt.h
+++ b/include/session/multi_encrypt.h
@@ -1,0 +1,140 @@
+#pragma once
+
+#ifdef __cplusplus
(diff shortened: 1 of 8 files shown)


jagerman:
Merging this as it's being used without any apparent issues now in dev versions of Session.

PR #78, opened 2024-02-02
Unified state

mpretty-cyro:
The goal of this PR is to consolidate the different config object types behind a single `State` class so clients only need to hold on to a single object instead of juggling multiple `Config` objects.

This PR also attempts to abstract as much of the config change/update loop as it currently can via generic hooks (`on_send`/`on_store`) and bulk processing methods (`state.merge()`).

In the future this class can be extended to include individual functions which result in changes across multiple configs in order to simplify the interface for the clients (eg. adding a member to a group).


include/session/config/namespaces.hpp
```diff
@@ -39,6 +39,8 @@ namespace {
 
             case Namespace::RevokedRetrievableGroupMessages:
                 return "RevokedRetrievableGroupMessages";
+
+            default: return "Invalid";
```
jagerman: Style thing: it would be nicer to put this `return` after the end of the switch instead of in a `default` case, because that way this whole switch would start producing warnings for unhandled cases if new enum values get added that aren't handled here, which is a nice reminder.  (With the `default` that won't happen).

mpretty-cyro: Updated

mpretty-cyro:
Converting this back to a draft as it'll require a number of changes to be usable again

PR #80, opened 2024-03-04
Length checking values

mpretty-cyro:
- Made length checks in explicit setters throw
- Truncating contact name (a contact's name should be based on that users `UserProfile->name` value which is length checked)

### DIFF
--- a/src/config/contacts.cpp
+++ b/src/config/contacts.cpp
@@ -41,7 +41,7 @@ contact_info::contact_info(std::string sid) : session_id{std::move(sid)} {
 
 void contact_info::set_name(std::string n) {
     if (n.size() > MAX_NAME_LENGTH)
-        throw std::invalid_argument{"Invalid contact name: exceeds maximum length"};
+        n = n.substr(0, MAX_NAME_LENGTH);
     name = std::move(n);
 }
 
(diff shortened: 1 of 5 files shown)


src/config/contacts.cpp
```diff
@@ -41,7 +41,7 @@ contact_info::contact_info(std::string sid) : session_id{std::move(sid)} {
 
 void contact_info::set_name(std::string n) {
     if (n.size() > MAX_NAME_LENGTH)
-        throw std::invalid_argument{"Invalid contact name: exceeds maximum length"};
+        n = n.substr(0, MAX_NAME_LENGTH);
```
jagerman:
```suggestion
        n.resize(MAX_NAME_LENGTH);
```
is slightly more efficient.

But also, is there a reason we *don't* want this one to throw if given something too long?

mpretty-cyro: Yep, a user never sets a contact name directly via the UI - a contacts name is included in the profile info included in messages so this is mostly to prevent `libSession` from throwing if another user has a modified client which sends an excessively long display name (doing this would probably crash Android at the moment 🤔)

jagerman: Ah, okay.

jagerman (approved):

PR #81, opened 2024-03-14
Removed the default rekey from the Keys init

mpretty-cyro:
There is a fun behaviour where when one device creates a group and adds members it generates an initial key (containing the members) but when a linked device creates an empty group state (pending the first poll) it ends up generating a new keys with no members.

Since the key used for encryption is based on the created timestamp this second "empty" key ends up incorrectly getting used.

This PR removes the default `rekey` that was called when initialising the `Keys` object, devs will need to manually call it when they are ready to create the initial key (this way the linked device won't create an invalid "empty" key).

### DIFF
--- a/src/config/groups/keys.cpp
+++ b/src/config/groups/keys.cpp
@@ -59,8 +59,6 @@ Keys::Keys(
         auto key_list = group_keys();
         members.replace_keys(key_list, /*dirty=*/false);
         info.replace_keys(key_list, /*dirty=*/false);
-    } else if (admin()) {
-        rekey(info, members);
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/config/groups/keys.cpp
```diff
@@ -59,8 +59,6 @@ Keys::Keys(
         auto key_list = group_keys();
         members.replace_keys(key_list, /*dirty=*/false);
         info.replace_keys(key_list, /*dirty=*/false);
-    } else if (admin()) {
-        rekey(info, members);
```
jagerman, line 63: The comment for the `rekey()` function in the header needs an update with this change about when it should be called, something along the lines of "This must also be called when creating a new group from scratch (but not when merely creating a new group *object* from new incoming group info such as from config data pushed by another device)."

PR #82, opened 2024-03-15
Add static submodule libs to bundled libsession-util

jagerman:
This automatically adds anything that we include via system_or_submodule into the libsession-util bundled library, most notably included libquic (in other upcoming PRs).

It also updates the libsession_static_bundle function so that only static library targets are actually bundled so that passing things like oxenc::oxenc (which is header-only and so not a static lib) just get silently ignored.

### DIFF
--- a/cmake/AddStaticBundleLib.cmake
+++ b/cmake/AddStaticBundleLib.cmake
@@ -1,6 +1,11 @@
 
 set(LIBSESSION_STATIC_BUNDLE_LIBS "" CACHE INTERNAL "list of libs to go into the static bundle lib")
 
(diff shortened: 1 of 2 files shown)


mpretty-cyro (approved):

jagerman:
I *think* this PR isn't needed anymore (after #83), because that bundling code in #83 just recursively follows target_link_libraries to pull everything in.

mpretty-cyro:
> I _think_ this PR isn't needed anymore (after #83), because that bundling code in #83 just recursively follows target_link_libraries to pull everything in.

Yep everything that was in here should be in #83

PR #83, opened 2024-04-03
Libquic integration

mpretty-cyro:


include/session/network.h
```diff
@@ -0,0 +1,257 @@
...
+        size_t body_size,
+        void (*callback)(
+                bool success,
+                bool timeout,
+                int16_t status_code,
+                const char* response,
+                size_t response_size,
+                network_service_node_changes changes,
+                void*),
+        void* ctx);
```
jagerman, line 347: 16 arguments is pretty hairy.  A nicer approach for the C API when arguments get this long might be to use a single struct that gets passed, along the lines of:

```C
// Callback function pointer used for all of the send-request functions:
typedef void (network_response_callback)(
                bool success,
                bool timeout,
                int16_t status_code,
                const char* response,
                size_t response_size,
                network_service_node_changes changes,
                void*);

typedef struct network_send_request {
        network_object* network;
        const char* method;
        const char* protocol;
        const char* host;
        const char* endpoint;
        uint16_t port;
        const char* x25519_pubkey;
        const char** query_param_keys;
        const char** query_param_values;
        size_t query_params_size;
        const char** headers;
        const char** header_values;
        size_t headers_size;
        const unsigned char* body;
        size_t body_size;
        network_response_callback* callback;
        void* ctx;
} network_send_request;

LIBSESSION_EXPORT void network_send_onion_request_to_server_destination(
    const network_send_request* request);
```
with application calling code that looks like this:
```C
network_send_request req = {0}; // initialize everything to zero/NULL
req.network = mynet;
req.method = "GET";
// etc.
// but don't need to set query_param_keys, query_param_values if you don't pass them since they are defaulted to null by the = {0};
// and then once all set:
network_send_onion_request_to_server_destination(&req);
```

which ought to make application code a bit easier to write.

include/session/network.hpp
```diff
@@ -0,0 +1,217 @@
...
+#include "session/types.hpp"
+
+namespace session::network {
+
+enum class ServiceNodeChangeType {
+    none = 0,
+    invalid_path = 1,
+    replace_swarm = 2,
+    update_path = 3,
+    update_node = 4,
```
jagerman: ~~The `= N` here is unnecessary (I think).  It's normal to leave them off if you don't care about the specific numeric value (that is: if there is never a need to cast these scoped enum values to/from a primitive integer).~~

jagerman: Ah crap, the value matters for interoperability with the C API. :(

include/session/network.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+struct onion_path {
+    std::shared_ptr<oxen::quic::connection_interface> conn;
+    std::vector<session::network::service_node> nodes;
+    uint8_t failure_count;
+};
+
+struct service_node_changes {
+    ServiceNodeChangeType type = ServiceNodeChangeType::none;
+    std::vector<session::network::service_node> nodes = {};
```
jagerman: The `= {};` is entirely optional for non-primitive types like std::vector.  (Basically anything with explicit constructors has to be constructed any time a containing object is constructed: only primitives like ints/bools are allowed to be in an uninitialized state).  It's perfectly fine to leave it if you want; it's just a style thing.

include/session/network.hpp
```diff
@@ -0,0 +1,217 @@
...
+    bool path_invalid = false;
+};
+
+struct request_info {
+    const service_node target;
+    const std::string endpoint;
+    const std::optional<ustring> body;
+    const std::optional<std::vector<service_node>> swarm;
+    const std::optional<onion_path> path;
+    const bool is_retry;
```
jagerman: I don't think these members should be `const`: if you already have a `const request_info` then all of its members are (to you) `const` already, and if you have a *non-*const on hand then everything being `const` just makes it more cumbersome to work with.

include/session/network.hpp
```diff
@@ -0,0 +1,217 @@
...
+    std::shared_ptr<oxen::quic::GNUTLSCreds> creds;
+    std::vector<onion_path> paths;
+
+    std::shared_ptr<oxen::quic::Endpoint> endpoint;
+    std::shared_ptr<oxen::log::RingBufferSink> buffer;
+
+  public:
+    // Constructs a new network for the given credentials, all requests should be made via a single
+    // Network instance.
+    Network(const session::onionreq::ed25519_seckey ed25519_seckey);
```
jagerman: As a general rule of thumb, you want to pass most arguments as either const *reference*, or non-const value:
```C++
void f(const SomeType& x);
void g(SomeType x);
```
The reason being that a function that takes a *const value* (no `&`), such as:
```C++
void h(const SomeType x)
```
forces the compiler to make a copy of SomeType into a local SomeType at the beginning of the function, but then makes that copy `const` for the duration of the function `h`.  (This isn't very useful because you can almost never do any more with a `const SomeType x` than you could with a `const SomeType& x` reference that it can just pass by pointer, but you forced the compiler to make a copy anyway).

And so the general rule of thumb is:
- pass by value (non-const!) if you need to make a copy, for example because you are going to mutate that local copy or store it yourself beyond just the duration of the function; and
- pass by const reference if you only need to reference it for the duration of the function and don't need to change or copy it.

(As an aside, there are also three other common argument passing types for other cases:
- `SomeType& x` -- called an lvalue reference -- where you want to directly modify the caller's variable.  Sometimes called an "output parameter" or "in/out" parameter, though that's a mildly misleading name.
- `SomeType&& x` -- called an rvalue reference -- where SomeType is a concrete type, this *forces* the caller to pass a reference to a temporary value into the function.  They can create a temporary on the spot (i.e. inside the argument), or explicitly `std::move(arg)` the argument, but they are *forbidden* from otherwise providing a reference to a variable they might use again.  This is a bit rare because for most movable types it's more convenient to just take a plain `SomeType x` by value, which allows the caller to decide if they want to move (by passing a temporary or a `std::move(arg)`) or copy (by passing a non-temporary value, which then copies it for the called function).
- A templated T&&x, such as `template <typename T> void f(T&& x) { ... }` -- this is called a "universal reference" and will be either a temporary or a reference or a const reference based on what the caller called it with.  This is mostly used with "perfect forwarding" and `std::forward` to pass references along without changing their reference type, which looks like:
    ```C++
    template <typename... Args>
    void f(Args&&... args) {
        g(std::forward<Args>(x)...);
    }
    ```
    which looks a bit cryptic, but is relatively common in code bases: essentially this calls `g(...)` with exactly the same arguments (including their reference type!) as `f(...)` was called with.  So if `g` took some things by reference, by value, and by rvalue, someone calling `f(...)` would have the same requirements on `f` as `g` imposes and it would fail to compile if called with arguments that `g()` can't be called with.
- There's technically a fourth `const SomeType&& x` but I've never seen it used because its meaning "take a reference to a temporary but make it a const reference" doesn't have much practical use.

)

jagerman: (There are quite a lot of these `const Type x` instead of `const Type& x` parameters, so I'll just leave this comment here for all of them that should just be changed to a reference type, and only note if I find one that should become something else).

include/session/network.hpp
jagerman: Okay I read further and see what's going on here: I think this struct is meant to be a container of references, rather than values, in which case it should be `const whatever& member`.

But I think this won't be sufficient because of other changes along the way, and so I think we need to rework this a little.

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+                if (resp.is_error() && !body.empty())
+                    throw std::runtime_error{"Failed to fetch response with error: " + body};
+                else if (resp.is_error())
+                    throw std::runtime_error{"Failed to fetch response"};
+
+                prom.set_value(body);
+            } catch (...) {
+                prom.set_exception(std::current_exception());
+            }
+        });
```
jagerman: The promise/future here makes `send_request` synchronous, meaning it will block the entire thread until the promise is fulfilled.

We need this to be more asynchronous, something more like this:

That means we need this command lambda to be a little different:
```C++
        get_btstream(info.target)->command(info.endpoint, payload, [resp_cb = std::move(handle_response)](message resp) {
            if (resp.timed_out)
                resp_cb(/*call with something here than indicates the timeout*/);

            std::string body = resp.body_str();
            if (resp.is_error()) {
                resp_cb(
                    /*call with something here that indicates we got an error response,
                       with an optional resp.body() if available); */
            }
            
            resp_cb(std::move(body));
        });
```

Otherwise the only option the application has is to give up control of whatever thread it is currently on until the response comes back from the other side, which doesn't work in general.

(And if a caller really does want to block -- like maybe in a command-line application to query a request, think along the lines of `curl` -- then it can pretty easily use a promise/future inside the callback itself to do so).

include/session/onionreq/builder.hpp
```diff
@@ -3,10 +3,46 @@
...
 #include <string_view>
 
+#include "../network_service_node.hpp"
 #include "key_types.hpp"
 
 namespace session::onionreq {
 
+struct SnodeDestination {
+    session::network::service_node node;
+    std::optional<std::vector<session::network::service_node>> swarm;
```
jagerman: Does this need to be optional, or could you just use an empty vector for the not-set state?

include/session/onionreq/builder.hpp
```diff
@@ -3,10 +3,46 @@
...
+
+    ServerDestination(
+            std::string protocol,
+            std::string host,
+            std::string endpoint,
+            session::onionreq::x25519_pubkey x25519_pubkey,
+            std::string method = "GET",
+            std::optional<uint16_t> port = std::nullopt,
+            std::optional<std::vector<std::pair<std::string, std::string>>> headers = std::nullopt,
+            std::optional<std::vector<std::pair<std::string, std::string>>> query_params =
```
jagerman: Isn't this customarily passed in the `endpoint` in a URL?  Is there a reason we need to pass it along separately?

include/session/onionreq/builder.hpp
```diff
@@ -75,8 +90,12 @@ class Builder {
 
     std::optional<std::string> host_ = std::nullopt;
     std::optional<std::string> target_ = std::nullopt;
+    std::optional<std::string> endpoint_ = std::nullopt;
     std::optional<std::string> protocol_ = std::nullopt;
+    std::optional<std::string> method_ = std::nullopt;
     std::optional<uint16_t> port_ = std::nullopt;
+    std::optional<std::vector<std::pair<std::string, std::string>>> headers_ = std::nullopt;
+    std::optional<std::vector<std::pair<std::string, std::string>>> query_params_ = std::nullopt;
```
jagerman, line 95: These don't need to be ` = std::nullopt` as that's the default constructed value of an optional anyway.

include/session/util.hpp
```diff
@@ -80,7 +80,7 @@ inline bool string_iequal(std::string_view s1, std::string_view s2) {
 
 // C++20 starts_/ends_with backport
 inline constexpr bool starts_with(std::string_view str, std::string_view prefix) {
-    return str.size() >= prefix.size() && str.substr(prefix.size()) == prefix;
+    return str.size() >= prefix.size() && str.substr(0, prefix.size()) == prefix;
 }
```
jagerman: We should drop these now that we have switched on C++20 compilation and just use `str.starts_with(prefix)` instead.

jagerman: And likewise for end_with below.

src/onionreq/builder.cpp
```diff
@@ -43,6 +44,92 @@ EncryptType parse_enc_type(std::string_view enc_type) {
...
+    // If we have query parameters, add them to the endpoint to be included in the payload
+    if (query_params_ && !query_params_->empty()) {
+        std::string query_string = "";
+
+        for (auto& query : *query_params_)
+            query_string += query.first + "=" + query.second + "&";
+
+        // Drop the extra '&' from the end
+        endpoint += "?" + query_string.substr(0, query_string.size() - 1);
+    }
```
jagerman: It feels a bit weird for Builder to be doing this query string reconstruction, rather than just requiring the application to pass it in as part of the endpoint.  (Plus that lets us completely avoid the question of how values should be properly escaped -- e.g. if someone gives a query parameter pair of `{"a", "="}` we really should be appending `a=%3d` instead of `a==`, but now that's leading into a rabbit hole of needing to know about proper URL encoding).

jagerman: Plus it would simplify a few function calls that are passing these in separately right now that could forget about them (i.e. and make the application worry about query parameters).

src/onionreq/builder.cpp
```diff
@@ -43,6 +44,92 @@ EncryptType parse_enc_type(std::string_view enc_type) {
...
+
+    // Structure the request information
+    nlohmann::json request_info{
+            {"method", *method_}, {"endpoint", endpoint}, {"headers", headers_json}};
+    auto request_info_dump = request_info.dump();
+    std::vector<std::string> payload{request_info_dump};
+
+    // If we were given a body, add it to the payload
+    if (body.has_value())
+        payload.emplace_back(std::string(from_unsigned(body->data()), body->size()));
```
jagerman: There's a `ustring_view` version of `from_unsigned` called `from_unsigned_sv` that makes this simpler:
```suggestion
        payload.emplace_back(from_unsigned_sv(*body));
```

src/onionreq/builder.cpp
```diff
@@ -43,6 +44,92 @@ EncryptType parse_enc_type(std::string_view enc_type) {
...
+
+        // Drop the extra '&' from the end
+        endpoint += "?" + query_string.substr(0, query_string.size() - 1);
+    }
+
+    // Structure the request information
+    nlohmann::json request_info{
+            {"method", *method_}, {"endpoint", endpoint}, {"headers", headers_json}};
+    auto request_info_dump = request_info.dump();
+    std::vector<std::string> payload{request_info_dump};
```
jagerman:
```suggestion
    std::vector<std::string> payload{std::move(request_info_dump)};
```
so that we don't need to make a copy of it.

Or alternatively just drop the temporary variable (and accomplishing the same optimization):
```suggestion
    std::vector<std::string> payload{request_info.dump()};
```

include/session/network_service_node.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "onionreq/key_types.hpp"
+
+namespace session::network {
+
+struct service_node {
+    std::array<uint8_t, 4> ip;
+    uint16_t quic_port;
+    session::onionreq::x25519_pubkey x25519_pubkey;
+    session::onionreq::ed25519_pubkey ed25519_pubkey;
```
jagerman: It might be worth storing a single `quic::RemoteAddress` in place of ip+port+ed25519_pubkey.  It offers a bunch of convenience methods, can do string conversion (from string on construction, to string via fmt) and is more simply passed into quic to establish a connection without needing to format the IP.

jagerman: (It doesn't store or need the x25519 pubkey, so that'll need to be separate if needed—it may not be needed at all).

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+                throw "Invalid malformed IPv4 address";
+        }
+
+        return quad;
+    }
+}  // namespace
+
+Network::Network(const session::onionreq::ed25519_seckey ed25519_seckey) {
+    creds = GNUTLSCreds::make_from_ed_seckey(std::string(ed25519_seckey.view()));
+    endpoint = net.endpoint(Address{"0.0.0.0", 0}, opt::outbound_alpns{{uALPN}});
```
jagerman:
```suggestion
    endpoint = net.endpoint(Address{"0.0.0.0", 0}, opt::alpns{{uALPN}});
```
Not strictly needed, but marginally less verbose.

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+                ec != std::errc{} || p != end)
+                throw "Invalid malformed IPv4 address";
+        }
+
+        return quad;
+    }
+}  // namespace
+
+Network::Network(const session::onionreq::ed25519_seckey ed25519_seckey) {
+    creds = GNUTLSCreds::make_from_ed_seckey(std::string(ed25519_seckey.view()));
```
jagerman: This secret key is the one that gets used for connections you make, but one consequence of this is that its corresponding pubkey identifies you (especially if you pass in your actual Session secret key here).

I think it would make more sense for us (in libsession) to simply use randomly generated Ed25519 pubkey for each connection we make, rather than holding onto a `creds` object.  (I'm thinking it might make sense for libquic to do this itself, eventually, as it's a fairly common thing for clients to want to do).  Ed25519 key generation is cheap, so it shouldn't be a performance issue to just do it on the fly.

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+    endpoint = net.endpoint(Address{"0.0.0.0", 0}, opt::outbound_alpns{{uALPN}});
+}
+
+void Network::add_logger(std::function<void(const std::string&)> callback) {
+    buffer = std::make_shared<oxen::log::RingBufferSink>(100UL, callback);
+    oxen::log::add_sink(buffer);
```
jagerman: If I remember correctly, this RingBufferSink keeps a buffer of the last N (100 here) log statements for later retrieval.  Is that wanted here?  If not (if all that is needed is the callback), there's a `spdlog::sinks::callback_sink` that might work better.

(The motivation for the RingBufferSink came from Lokinet, where we wanted a frontend GUI client to be able to connect to the backend actual running lokinet and retrieve recent log statements).

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+            throw "Invalid IPv4 address";
+        for (int i = 0; i < 4; i++) {
+            auto end = nums[i].data() + nums[i].size();
+            if (auto [p, ec] = std::from_chars(nums[i].data(), end, quad[i]);
+                ec != std::errc{} || p != end)
+                throw "Invalid malformed IPv4 address";
+        }
+
+        return quad;
+    }
```
jagerman, line 293: With my other suggestion about using quic::RemoteAddress (or possibly quic::Address) we can get rid of this manual IP splitting logic and just use quic::Address construction (which will throw if given an invalid IP).

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+            return result.first;
+
+        // Otherwise if there was no connection at all then return nullptr immediately so a new
+        // connection gets created (this can happen if the `conn` times out or fails initially so we
+        // don't want to bother looping in that case)
+        if (!result.second)
+            break;
+
+        std::this_thread::sleep_for(100ms);
+        wait_time_ms += 100ms;
```
jagerman: This sleep/timer shouldn't be required: the connection on our side will exist right away.  But I think what is happening is that the `maybe_stream` is returning false because Stream creation works slightly differently from Connection creation: the stream doesn't technically exist until the remote has allowed us to open streams.

We *can* create a Stream object, and have it locally on our side, it just won't be connected yet (and I think that's why `maybe_stream(0)` isn't returning it -- because technically it doesn't get that stream id assigned until its allowed to create a stream).  Instead it gets activated automatically as soon as the remote signals at the end of the handshake that we may now open streams, but the Stream object we create can and does exist before that, just in a "pending" state.

The solution, then, is that I was wrong in my initial suggestion here: rather than just storing the Connection and letting the stream live inside it, we need to store both side-by-side (e.g. in a `std::pair<Connection, BTRequestStream>` or maybe nicer in a `struct`), which should deal with that case.  When we create, with `connect` and then immediately make a BTRequestStream on that connection (it will be pending, but will still be a usable stream, from the application's perspective).  If we already in the midst of opening a connection and establishing the initial stream, we'll then get both and can start making requests on the stream right away.  (As for what happens when you start sending on a pending stream: we just buffer the data/requests and fire them off as soon as the stream leaves pending status).

jagerman: I think there's also a little window here where we might double-establish a connection and then drop one, and perhaps we should solve that by making `get_btstream` both check and attempt to establish (if not already found) all inside the `call_get` lambda.  In other words, by moving the code just below this comment into the above `call_get` to have a single, larger call_get that both checks for an existing one and, if not found, makes a new one.  The mutex-like behaviour of call_get ought to rule out any double-connections, and combined with the putting Conns and Streams together, should resolve the weirdness you were seeing.

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+
+        if (target_path == paths.end())
+            return;
+
+        target_path->conn = std::move(c);
+    });
+    std::shared_ptr<oxen::quic::BTRequestStream> str =
+            c->open_stream<oxen::quic::BTRequestStream>();
+
+    return str;
```
jagerman: Marginally simpler:
```suggestion
    return c->open_stream<oxen::quic::BTRequestStream>();
```

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+                                [&destination](const service_node& node) {
+                                    return node == destination.node;
+                                });
+                    }),
+            valid_paths.end());
+    return valid_paths;
+}
+
+template <typename Destination>
+void Network::send_onion_request(
```
jagerman: This template use isn't quite right: the template itself is declared in the header, but is only actually defined here in the .cpp file.  But a template has to get compiled, which means any *other* code that attempts to call the templated `send_onion_request` will fail to compile as the template isn't available.  You can have templates in .cpp files, but generally only if the template is only used within the .cpp file itself.

The best way out here is probably something along the lines of:

```C++
using network_destination = std::variant<SnodeDestination, SomeOtherDestination, /*...*/>
```
and take a `network_destination destination` here.  (That will cascade into a few places, but I think those cascading locations have a similar issue).

to change `destination` to a `std::variant<SnodeDestination>`

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+        auto builder = Builder();
+        builder.set_destination(destination);
+
+        for (const auto& node : path.nodes)
+            builder.add_hop({node.ed25519_pubkey, node.x25519_pubkey});
+
+        auto payload = builder.generate_payload(body);
+        auto onion_req_payload = builder.build(payload);
+
+        request_info info = {
```
jagerman: The `=` here is redundant (i.e. `SomeType x{constructor, arguments}` and `SomeType x = {constructor, arguments};` do the same thing.  (This is just a style point; feel free to keep the ` = ` if you prefer it).

src/network.cpp
```diff
@@ -0,0 +1,1106 @@
...
+                // If there is no response data or no swarm informaiton was provided then we should
+                // just replace the swarm
+                if (!info.swarm)
+                    throw std::invalid_argument{"Unable to handle redirect."};
+
+                // If this was the first 421 then we want to retry using another node in the swarm
+                // to get confirmation that we should switch to a different swarm
+                if (!info.is_retry) {
+                    std::random_device rd;
+                    std::mt19937 g(rd());
```
jagerman: Going out to the system random device isn't all that desirable (it can, for instance, block if called too much).  Too alternative approaches:
- copy CSRNG out of lokinet (from crypto/crypto.hpp).  Then you can just use it as the RNG instead of `mt19937`, i.e.:
```C++
CSRNG rng;
std::shuffle(swarm_copy.begin(), swarm_copy.end(), rng);
```
(and you get cryptographically secure random via libsodium, though that isn't needed here).

Option 2:
Make a global, thread_local `std::mt19937_64` that gets seeded once from `std::random_device` in its construction, then just use that global RNG here (and anywhere else that needs a non-secure random) instead of `g`.  oxen-core has an example of doing this in src/common/random.{h,cpp}.

include/session/network.hpp
jagerman: > But I think this won't be sufficient because of other changes along the way, and so I think we need to rework this a little.

To elaborate: I left a comment here about not using a promise/future, and I think to make that work we'll need to make a copy of this object.

src/network.cpp
mpretty-cyro: Ah no worries, I wasn't sure if we should be using the users secret key here or if a random one was fine - making this always random will simplify the integration quite nicely 😁

src/network.cpp
mpretty-cyro: Updated

src/network.cpp
mpretty-cyro: Ah yep, I remember you mentioning this - it must have just slipped my mind when I was plugging it in (I think I might have copied the `RingBufferSink` from someones code in Discord)

src/network.cpp
mpretty-cyro: Yep good point - I've removed the promise and just trigger the original callback within the `command` callback as shown here.

src/network.cpp
mpretty-cyro: I ended up pulling in `CSRNG` and going with that approach as I needed to add a few more `shuffle` calls after pulling in the snode pool and path building logic so figured it may as well all be secure and consistent

src/onionreq/builder.cpp
mpretty-cyro: Oh nice - I didn't realise that it would automatically convert the `std::string_view` to a `std::string` for the `std::vector<std::string> payload` 👍

mpretty-cyro: Have made this change

src/onionreq/builder.cpp
mpretty-cyro: Ah yep, I think the temporary variable was left over from when I was debugging this initially 😞 

Have removed it

src/network.cpp
mpretty-cyro: I tried this but when I did `if constexpr (std::is_same_v<T, SnodeDestination>) {` inside `var::visit(` it was always returning false (whereas if I `if (auto* dest = std::get_if<SnodeDestination>(&destination)) {` it's working correctly) - not sure if this is iOS doing weird stuff (I did see that there are MacOS work-arounds so not sure if they need to also apply to iOS 🤷

src/onionreq/builder.cpp
mpretty-cyro: My original rationale was that it would mean the clients can avoid doing this construction logic (though, now that I think about it, each platform should have some native way to do it anyway)

I'll change this to just take the fully constructed endpoint

include/session/onionreq/builder.hpp
mpretty-cyro: No real reason - iOS was doing this manually so I just pulled the code in, i'll tweak the logic to use native methods and expect the fully constructed endpoint to be passed in instead

src/network.cpp
mpretty-cyro: Have update this

src/network.cpp
mpretty-cyro: Have reworked this now so it's no longer using this sleep/timer

include/session/network.hpp
mpretty-cyro: I've moved ownership of the service node cache into `libSession` which means this `service_node_changes` has been removed (as we can just apply changes to the internal cache now)

include/session/network.h
mpretty-cyro: I've moved the majority of these parameters into a `network_server_destination` so now `network_send_onion_request_to_snode_destination` and `network_send_onion_request_to_server_destination` have the same structure except for the `node`/`server` parameter

include/session/onionreq/builder.hpp
mpretty-cyro: Now that the snode cache is internal this has been changed to a `std::optional<std::string> swarm_pubkey` (which is mostly needed because you can send a request directly to a node unrelated to a swarm - if `swarm_pubkey` is assigned and we get a `421` error then the cached swarm info will be updated based on the error handling logic)

src/network.cpp
mpretty-cyro: Have updated this to avoid creating the local variable

src/network.cpp
mpretty-cyro: Updated

include/session/network.hpp
mpretty-cyro: I've removed the `const` from these - now that the snode pool is internal it might be easier to see how this is used (it's essentially only used to allow for the `retry` request in the case of receiving a `421`)

include/session/network.hpp
mpretty-cyro: I've gone through and cleaned these all up - I have left a few `const` parameters in the C API functions just because Swift interprets it differently with/without the `const` (ie. we need a mutable pointer if it's not a `const` variable)

include/session/network_service_node.hpp
mpretty-cyro: Managed to get this working so now it uses the `quic::RemoteAddress` and derives the `x25519` pubkey from the `ed25519` pubkey when needed

src/network.cpp
mpretty-cyro: This is now removed

src/network.cpp
jagerman: Sorry for not noticing this earlier: what is `T` here?  Inside a var::visit generic lambda taking a `auto& x` argument, you'd need something along the lines of this to extract the type:
```C++
using T = std::remove_cvref_t<decltype(x)>;
```
to remove any reference and/or const from the type  ("cvref" = const, volatile, reference; volatile is rare but const and reference are quite common), to get the plain, unadorned underlying type.

But since we now are under C++20, using a templated lambda makes this easier:
```C++
[]<typename T>(const T& x) { ... }
```
(so that the `const` and `&` are "outside" the T and thus T will be a plain type).

mpretty-cyro: All good - I don't have the original code anymore but it should have been something like this:
```
var::visit([this](const auto& dest){
    using T = std::remove_cvref_t<decltype(dest)>;
    if constexpr (std::is_same_v<T, oxen::quic::RemoteAddress>) 
        ed25519_public_key_.emplace(ed25519_pubkey::from_bytes(dest.view_remote_key()));
    else if constexpr (std::is_same_v<T, ServerDestination>) {
        host_.emplace(dest.host);
        endpoint_.emplace(dest.endpoint);
        protocol_.emplace(dest.protocol);
        method_.emplace(dest.method);

        if (dest.port)
            port_.emplace(*dest.port);

        if (dest.headers)
            headers_.emplace(*dest.headers);
    } else
        throw std::invalid_argument{"Invalid destination type."};
}, destination);
```

I think it was previously always throwing the `Invalid destination type` error, but for some reason it seems to be working now so I'm not sure what was going on...

Since reworking the logic I now only need to deal with this behaviour in the following 3 places:

```
void Builder::set_destination(network_destination destination) {
    ed25519_public_key_.reset();

    if (auto* dest = std::get_if<oxen::quic::RemoteAddress>(&destination))
        ed25519_public_key_.emplace(ed25519_pubkey::from_bytes(dest->view_remote_key()));
    else if (auto* dest = std::get_if<ServerDestination>(&destination)) {
        host_.emplace(dest->host);
        endpoint_.emplace(dest->endpoint);
        protocol_.emplace(dest->protocol);
        method_.emplace(dest->method);

        if (dest->port)
            port_.emplace(*dest->port);

        if (dest->headers)
            headers_.emplace(*dest->headers);
    } else
        throw std::invalid_argument{"Invalid destination type."};
}

std::optional<service_node> node_for_destination(network_destination destination) {
    if (auto* dest = std::get_if<oxen::quic::RemoteAddress>(&destination))
        return *dest;

    return std::nullopt;
}

session::onionreq::x25519_pubkey pubkey_for_destination(network_destination destination) {
    if (auto* dest = std::get_if<oxen::quic::RemoteAddress>(&destination))
        return compute_xpk(dest->view_remote_key());

    if (auto* dest = std::get_if<ServerDestination>(&destination))
        return dest->x25519_pubkey;

    throw std::runtime_error{"Invalid destination."};
}
```

include/session/network.hpp
```diff
@@ -0,0 +1,421 @@
...
+    ///
+    /// A recursive function that sends a request to provided nodes until a successful response is
+    /// received or the list is drained.
+    ///
+    /// Inputs:
+    /// - `target_nodes` -- [in] list of nodes to send requests to until we get a result or it's
+    /// drained.
+    /// - `callback` -- [in] callback to be triggered once we make a successful request.  NOTE: If
+    /// we drain the `target_nodes` and haven't gotten a successful response a NULL
+    /// `connection_info` and empty will be provided.
```
jagerman: I think this got copied from the C API, but I'm not sure what it should be for the C++ API.

include/session/onionreq/builder.hpp
```diff
@@ -1,12 +1,42 @@
...
+    std::optional<std::vector<std::pair<std::string, std::string>>> headers;
+    std::string method;
+
+    ServerDestination(
+            std::string protocol,
+            std::string host,
+            std::string endpoint,
+            session::onionreq::x25519_pubkey x25519_pubkey,
+            std::optional<uint16_t> port,
+            std::optional<std::vector<std::pair<std::string, std::string>>> headers,
```
jagerman: Perhaps these two should get ` = std::nullopt` defaults?

src/config/base.cpp
```diff
@@ -203,8 +204,8 @@ std::vector<std::string> ConfigBase::_merge(
     //   might be our current config, or might be one single one of the new incoming messages).
     // - confs that failed to parse (we can't understand them, so leave them behind as they may be
     //   some future message).
-    int superconf = new_conf->unmerged_index();  // -1 if we had to merge
-    for (int i = 0; i < all_hashes.size(); i++) {
+    size_t superconf = new_conf->unmerged_index();  // -1 if we had to merge
```
jagerman: This is potentially problematic: `superconf` here uses -1 as a special had-to-merge value, but as a size_t that becomes a large positive integer.  I think we want to fix the signed-compare-mismatch here the other way, by using a `static_cast<int>(all_hashes.size())`.

jagerman: Actually this should probably be an optional<size_t> instead; I'm redoing it as such.

include/session/onionreq/builder.hpp
mpretty-cyro: These defaults have now been added

src/config/base.cpp
```diff
@@ -651,53 +651,65 @@ LIBSESSION_EXPORT config_string_list* config_merge(
...
+        confs.reserve(count);
+        for (size_t i = 0; i < count; i++)
+            confs.emplace_back(msg_hashes[i], ustring_view{configs[i], lengths[i]});
+
+        return make_string_list(config.merge(confs));
+    } catch (const std::exception& e) {
+        copy_c_str(conf->_error_buf, e.what());
+        conf->last_error = conf->_error_buf;
+    }
+    return nullptr;
```
jagerman: We could slightly genericize this error handling with a pair of templated functions like this:

```C++
// No return value: accepts void and pointer returns; pointer returns will become nullptr on error
template <std::invocable Call>
decltype(f()) exc_to_error(config_object* conf, Call& f) {
    using Ret = decltype(f());
    try {
        return f();
    } catch (const std::exception& e) {
        copy_c_str(conf->_error_buf, e.what());
        conf->last_error = conf->_error_buf;
    }
    if constexpr (std::is_pointer_v<Ret>)
        return nullptr;
    else static_assert(std::is_void_v<Ret>, "Don't know how to return an error value!");
}

// Accepts callbacks with value returns on errors: returns `f()` on success, `error_return` on exception
template <std::invocable Call, typename Ret>
Ret exc_to_error(config_object* conf, Call& f, Ret error_return) {
    try {
        return f();
    } catch (const std::exception& e) {
        copy_c_str(conf->_error_buf, e.what());
        conf->last_error = conf->_error_buf;
    }
    return error_return;
}
```

and then can replace all of these `try-stuff-catch-set-error` blocks with:
```C++
LIBSESSION_EXPORT config_string_list* config_merge(
        config_object* conf,
        const char** msg_hashes,
        const unsigned char** configs,
        const size_t* lengths,
        size_t count) {
    return exc_to_error(conf, [&] {
        auto& config = *unbox(conf);
        std::vector<std::pair<std::string, ustring_view>> confs;
        confs.reserve(count);
        for (size_t i = 0; i < count; i++)
            confs.emplace_back(msg_hashes[i], ustring_view{configs[i], lengths[i]});

        return make_string_list(config.merge(confs));
    });
```

It's not a *huge* gain, but it does reduce the repetitiveness a bit.

The first version of the function will work in most places (with `void` or pointer returns -- the latter returns nullptr on exception), but if there's some version that returns a value (say an int) then the second one lets you do:
```C++
return exc_to_error(conf, [&] { return get_some_integer_that_might_throw(); }, -1);
```
to specify the value to return if an exception happens.

jagerman: `exc_to_error` is probably not a great name.  Maybe `wrap_exceptions` or something?

mpretty-cyro: No worries, I've added these in and updated all of the C functions which call throwing C++ functions to use it

mpretty-cyro:
Replaced by https://github.com/session-foundation/libsession-util/pull/6

PR #86, opened 2024-06-17
Added code to handle unwrapping double-wrapped config messages

mpretty-cyro:
It looks like there was an Android release which resulted in double-wrapped config messages (I'm guessing the `libSession` submodule was updated past version `1.1` but the original wrapping code wasn't removed from the client which resulted in both the client, and `libSession`, wrapping the config message)

This PR tries to unwrap the config message twice just in case we receive a config message from the buggy client.

### DIFF
--- a/src/config/base.cpp
+++ b/src/config/base.cpp
@@ -75,7 +75,23 @@ std::vector<std::string> ConfigBase::merge(
                         ustring_view{_keys.front().data(), _keys.front().size()},
                         c,
                         storage_namespace());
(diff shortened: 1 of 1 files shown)


jagerman:
I was thinking about how you might clean this up with a loop inside a single try, but anything I came up with ended up not any nicer.

jagerman (approved):

mpretty-cyro:
Yea, I kind of came to the same conclusion - the code doesn't look nice regardless so hopefully it'll annoy us enough to push to remove it as soon as we can 😅

PR #87, opened 2024-06-17
Added functions for generating a blinded id for version check auth

mpretty-cyro:


src/blinding.cpp
```diff
@@ -298,6 +301,37 @@ std::pair<uc32, cleared_uc32> blind25_key_pair(
...
 
+static const auto version_blinding_hash_key_sig = to_unsigned_sv("VersionCheckKey_sig"sv);
+
+std::pair<uc32, cleared_uc64> blind_version_key_pair(ustring_view ed25519_sk) {
+    std::array<unsigned char, 64> ed_sk_tmp;
+    if (ed25519_sk.size() == 32) {
+        std::array<unsigned char, 32> pk_ignore;
+        crypto_sign_ed25519_seed_keypair(pk_ignore.data(), ed_sk_tmp.data(), ed25519_sk.data());
+        ed25519_sk = {ed_sk_tmp.data(), 64};
+    }
```
jagerman: This temp sk ought to go into a cleared_uc64, if needed:
```suggestion
    std::optional<cleared_uc64> ed_sk_tmp;
    if (ed25519_sk.size() == 32) {
        std::array<unsigned char, 32> pk_ignore;
        ed_sk_tmp.emplace();
        crypto_sign_ed25519_seed_keypair(pk_ignore.data(), ed_sk_tmp->data(), ed25519_sk.data());
        ed25519_sk = {ed_sk_tmp->data(), 64};
    }
```

src/blinding.cpp
```diff
@@ -298,6 +301,37 @@ std::pair<uc32, cleared_uc32> blind25_key_pair(
...
+    crypto_generichash_blake2b(
+            seed.data(),
+            seed.size(),
+            ed25519_sk.data(),
+            32,
+            version_blinding_hash_key_sig.data(),
+            version_blinding_hash_key_sig.size());
+
+    uc32 pk;
+    cleared_uc64 sk;
```
jagerman:
```suggestion
    std::pair<uc32, cleared_uc64> result;
    auto& [pk, sk] = result;
```
and then just `return result;` at the end of the func.

By constructing directly into a value of the function's return type, and then returning that value, C++ return value optimization kicks in and will avoid having a that return variable in this function at all: the compiler will write the returned value directly into wherever the calling function is going to store it.

src/blinding.cpp
jagerman: Actually I don't think this get-the-full-secret-key code is needed at all for this function.

src/blinding.cpp
```diff
@@ -298,6 +301,37 @@ std::pair<uc32, cleared_uc32> blind25_key_pair(
...
+    std::array<unsigned char, 64> ed_sk_tmp;
+    if (ed25519_sk.size() == 32) {
+        std::array<unsigned char, 32> pk_ignore;
+        crypto_sign_ed25519_seed_keypair(pk_ignore.data(), ed_sk_tmp.data(), ed25519_sk.data());
+        ed25519_sk = {ed_sk_tmp.data(), 64};
+    }
+    if (ed25519_sk.size() != 64)
+        throw std::invalid_argument{
+                "blind_version_key_pair: Invalid ed25519_sk is not the expected 32- or 64-byte "
+                "value"};
```
jagerman: Most of this is not needed: we really only just need a check for size == 32 or size == 64 (technically just >= 32 is enough, but it's probably worth erroring if we get some oddball size).    The reason it isn't needed is that `crypto_sign_ed25519_seed_keypair` doesn't do anything to the seed: it just copies it into the first 32-bytes of the result, then follows it by the 32-byte pubkey, but since we only use the 32-byte seed, this isn't something we care about and can just use the first 32 bytes of `ed25519_sk` as the seed.

src/blinding.cpp
```diff
@@ -298,6 +301,37 @@ std::pair<uc32, cleared_uc32> blind25_key_pair(
...
+
+    uc32 seed;
```
jagerman: This ought to be in a cleared_uc32, but actually we can optimize that a bit by moving the `cleared_uc64 sk;` from further down up here to do double-duty (and avoid some secure erases):

```C++
    std::pair<uc32, cleared_uc64> result;
    auto& [pk, sk] = result;

    crypto_generichash_blake2b(
            sk.data(),
            32,
            ed25519_sk.data(),
            /* ... */);
    
    // Reuse `sk` to avoid needing extra secure erasing:
    if (0 != crypto_sign_ed25519_seed_keypair(pk.data(), sk.data(), sk.data()))
```

src/blinding.cpp
jagerman: (Included this change in an earlier-in-the-file suggestion, but leaving this here to explain why the change to a std::pair usage).

src/blinding.cpp
```diff
@@ -298,6 +301,37 @@ std::pair<uc32, cleared_uc32> blind25_key_pair(
...
+            32,
+            version_blinding_hash_key_sig.data(),
+            version_blinding_hash_key_sig.size());
+
+    uc32 pk;
+    cleared_uc64 sk;
+    if (0 != crypto_sign_ed25519_seed_keypair(pk.data(), sk.data(), seed.data()))
+        throw std::runtime_error{"blind_version_key_pair: ed25519 generation from seed failed"};
+
+    return {pk, sk};
```
jagerman:
```suggestion
    return result;
```

src/blinding.cpp
```diff
@@ -492,6 +544,21 @@ LIBSESSION_C_API bool session_blind25_key_pair(
     }
 }
 
+LIBSESSION_C_API bool session_blind_version_key_pair(
+        const unsigned char* ed25519_seckey,
+        unsigned char* blinded_pk_out,
+        unsigned char* blinded_sk_out) {
+    try {
+        auto result = session::blind_version_key_pair({ed25519_seckey, 64});
+        auto [b_pk, b_sk] = result;
```
jagerman: As written, this is going to make a copy of the pk and sk values (and then a second copy in the `memcpy`).  Two ways around it:

Don't use the intermediate `result` and read them directly into b_pk/b_sk:
```suggestion
        auto [b_pk, b_sk] = session::blind_version_key_pair({ed25519_seckey, 64});
```
Or make b_pk, b_sk references into `result`:
```suggestion
        auto result = session::blind_version_key_pair({ed25519_seckey, 64});
        auto& [b_pk, b_sk] = result;
```

src/blinding.cpp
```diff
@@ -526,6 +593,22 @@ LIBSESSION_C_API bool session_blind25_sign(
...
 
+LIBSESSION_C_API bool session_blind_version_sign(
+        const unsigned char* ed25519_seckey,
+        CLIENT_PLATFORM platform,
+        size_t timestamp,
+        unsigned char* blinded_sig_out) {
+    try {
+        auto result = session::blind_version_sign(
+                {ed25519_seckey, 64}, static_cast<Platform>(platform), timestamp);
+        auto sig = result;
```
jagerman:
```suggestion
        auto sig = session::blind_version_sign(
                {ed25519_seckey, 64}, static_cast<Platform>(platform), timestamp);
```
(avoids unintentionally copy)

include/session/sodium_array.hpp
```diff
@@ -0,0 +1,231 @@
+#pragma once
```
jagerman, line 1: I'm a bit confused -- didn't we already have this header in here somewhere?

jagerman: Oh never mind, I see this was just moved away from util.hpp.

src/blinding.cpp
jagerman: That's a bit confusing.  The point is:

`crypto_sign_ed25519_seed_keypair([seed32])` returns `[seed32, pk32]` packed into a 64-byte location, and so because we are only using the first 32 bytes anyway, we don't need the `crypto_sign_ed25519_seed_keypair` call (which only really produces new data in the pk32 that we are ignoring).

mpretty-cyro: Ah yep, nice - I'll remove it

src/blinding.cpp
mpretty-cyro: I'm not sure how long it'll take me to get used to this `auto& [pk, sk] = result;` convention (it's just so different from most of the languages I normally work in 😅) but good to know that it also results in optimisations in the compiler when done

Have updated the code 🙏

src/blinding.cpp
mpretty-cyro: Nice - have updated it

src/blinding.cpp
mpretty-cyro: Updated

src/blinding.cpp
mpretty-cyro: Not sure why I did this - I guess I just copied it from `session_blind25_key_pair` and assumed it was needed... and it looks like I wrote those functions as well so no idea 😂 

I've updated all of them to avoid this now

src/blinding.cpp
mpretty-cyro: Similar to the above, looks like I just copied this from the blind_15/25 functions which I wrote a while back that was also doing this - have updated them as well

include/session/sodium_array.hpp
mpretty-cyro: Yea - I think we found this caused some linking issues in the `libQuic` branch - maybe something to do with the `ENABLE_ONIONREQ` flag? Or just trying to avoid a future conflict? I can't remember exactly 😞

They _should_ have also been removed from the `util.cpp` though which they hadn't been so I've cleaned that up

src/blinding.cpp
mpretty-cyro: So I've included this change but just wanted to flag that it ended up resulting in changing the output of the function (see unit test changes below), I don't really understand _why_ the result would have changed though because it seems to only use the first 32 bytes of `sk` 😕 

```

~/LibSession-Util/tests/test_blinding.cpp:282: FAILED:
  CHECK( oxenc::to_hex(seckey.begin(), seckey.end()) == "91faddc2c36da4f7bcf24fd977d9ca5346ae7489cfd43c58cad9eaaa6ed60f6988e8adb27e7b8ce776fcc25b" "c1501fb2888fcac0308e52fb10044f789ae1a8fa" )
with expansion:
  "4091bdaafefd7ddc3398db877c894716b2b24f12dac15ad414a3e4f0b6ac1c6788e8adb27e7-
  b8ce776fcc25bc1501fb2888fcac0308e52fb10044f789ae1a8fa"
  ==
  "91faddc2c36da4f7bcf24fd977d9ca5346ae7489cfd43c58cad9eaaa6ed60f6988e8adb27e7-
  b8ce776fcc25bc1501fb2888fcac0308e52fb10044f789ae1a8fa"

~/LibSession-Util/tests/test_blinding.cpp:291: FAILED:
  CHECK( oxenc::to_hex(signature.begin(), signature.end()) == "143c2c9828f7680ee81e6247bc7aa4777c4991add87cd724149b00452bed4e920fa57daf4627c68f43fcbddb" "2d465d5ea11def523f3befb2bbee39c769676305" )
with expansion:
  "adfb25eafec6fb0037fc39145e64badfbdd08cc95a77c01577ecc623fd856fb05b90a921fc6-
  b805e6a730ca9505e1c069f256020a76beb2ecbb6e32c47e12104"
  ==
  "143c2c9828f7680ee81e6247bc7aa4777c4991add87cd724149b00452bed4e920fa57daf462-
  7c68f43fcbddb2d465d5ea11def523f3befb2bbee39c769676305"

```

src/blinding.cpp
jagerman: It's a shortcut for `auto& pk = result.first; auto& sk = result.second;`, but works for std::pairs, std::array, std::tuples, and simple structs.  It's particularly useful in looping through maps:
```C++
for (auto& [k, v] : somemap) { /*...*/ }
```

src/blinding.cpp
jagerman: That seems like a problem.  I'll investigate.

jagerman: It seems my suggestion was wrong: libsodium does not allow the seed input and secret key output to overlap here.  I've pushed a fix to put the temp seed in a separate cleared_uc32 instead.

jagerman (approved):

src/blinding.cpp
mpretty-cyro: Ah yea that makes a bit more sense in my head haha

src/blinding.cpp
mpretty-cyro: Ah no worries, thanks 🙏

PR #89, opened 2024-06-19
fix: build for msvc

Bilb:


include/session/config/groups/members.h
```diff
@@ -55,7 +55,11 @@ LIBSESSION_EXPORT int groups_members_init(
         const unsigned char* ed25519_secretkey,
         const unsigned char* dump,
         size_t dumplen,
-        char* error) __attribute__((warn_unused_result));
+        char* error)
+#ifdef __GNUC__
+        __attribute__((warn_unused_result))
+#endif
+        ;
```
jagerman: There are a bunch more of these in other C headers.  It would be nicer to add this into `export.h` (which gets included from all the C headers):
```C++
#ifdef __GNUC__
#define LIBSESSION_WARN_UNUSED __attribute__((warn_unused_result))
#else
#define LIBSESSION_WARN_UNUSED
#endif
```
and then we can just `__attribute__((warn_unused_result));` -> `LIBSESSION_WARN_UNUSED;` everywhere rather than needing a few dozen ifdefs.

(I count a couple dozen in session/config/various*.h, and one in session/network.h)

src/config/groups/keys.cpp
```diff
@@ -1286,7 +1286,7 @@ std::pair<std::string, ustring> Keys::decrypt_message(ustring_view ciphertext) c
     //
     // Removing any null padding bytes from the end
     //
-    if (auto pos = plain.find_last_not_of('\0'); pos != std::string::npos)
+    if (auto pos = plain.find_last_not_of((unsigned char) 0); pos != std::string::npos)
```
jagerman: That should be able to be just a plain `0` (i.e. I don't think the explicit `(unsigned char)` cast is needed).  If compilers are warning about that, though, then `unsigned char{0}` is a slightly cleaner way to write it.

src/hash.cpp
```diff
@@ -31,7 +31,7 @@ ustring hash(const size_t size, ustring_view msg, std::optional<ustring_view> ke
     if (result_code != 0)
         throw std::runtime_error{"Hash generation failed"};
 
-    return {result, size};
+    return to_unsigned(static_cast<unsigned char*>(result));
```
jagerman: There's some weirdness in this function (including a [VLA](https://en.wikipedia.org/wiki/Variable-length_array)).  I've pushed some code to clean it up, which changes this around to something nicer that MSVC should be happy with.

src/session_encrypt.cpp
```diff
@@ -584,7 +584,7 @@ ustring decrypt_push_notification(ustring_view payload, ustring_view enc_key) {
         throw std::runtime_error{"Failed to decrypt; perhaps the secret key is invalid?"};
 
     // Removing any null padding bytes from the end
-    if (auto pos = buf.find_last_not_of('\0'); pos != std::string::npos)
+    if (auto pos = buf.find_last_not_of((unsigned char)0); pos != std::string::npos)
```
jagerman, line 587: (Same as earlier comment)

src/config/groups/keys.cpp
Bilb: Just tried making is just `0`, but I get 
```
D:\a\libsession-util-nodejs\libsession-util-nodejs\libsession-util\src\config\groups\keys.cpp(1289,26): error C2668: 'std::basic_string<unsigned char,std::char_traits<unsigned char>,std::allocator<unsigned char>>::find_last_not_of': ambiguous call to overloaded function [D:\a\libsession-util-nodejs\libsession-util-nodejs\build\libsession-util\src\config.vcxproj]

```
for all builds
https://github.com/oxen-io/libsession-util-nodejs/actions/runs/9591309454/job/26447994842

Bilb: I had to make it 
`(unsigned char){0}`
as I'd get `type  name is not allowed`
without the brackets

Bilb: actually, msvc doesn't like this, so reverting to the good ol' `unsigned char(0)`
https://github.com/oxen-io/libsession-util-nodejs/actions/runs/9591425021/job/26448313193

src/session_encrypt.cpp
Bilb: see comment above, that msvc doesn't like it

include/session/config/groups/members.h
Bilb: should be fixed (but I don't have a network.h on my current branch)

src/session_encrypt.cpp
jagerman: Shoot, well okay then.

include/session/config/groups/members.h
jagerman: Ah, right, that is from the libquic branch.

jagerman (approved):

PR #92, opened 2024-07-17
Additional invite/promotion state and missing C functions

mpretty-cyro:
- Added an 'INVITE_NOT_SENT' status which is the default used when creating a new group member
- Added a groups_keys_free function to the C API
- Added a groups_keys_storage_namespace function to the C API
- Updated the promotion functions based on the new status options

### DIFF
--- a/include/session/config/groups/members.hpp
+++ b/include/session/config/groups/members.hpp
@@ -81,14 +81,11 @@ struct member {
     ///
     /// Member variable
     ///
-    /// Flag that is set to indicate to the group that this member is an admin.
+    /// Flag that is set to indicate to the group that this member has been promoted to an admin.
     ///
     /// Note that this is only informative but isn't a permission gate: someone could still possess
     /// the admin keys without this (e.g. if they cleared the flag to appear invisible), or could
     /// have lost (or never had) the keys even if this is set.
-    ///
-    /// See also `promoted()` if you want to check for either an admin or someone being promoted to
-    /// admin.
     bool admin = false;
 
     /// API: groups/member::supplement
@@ -129,16 +126,16 @@ struct member {
         supplement = false;
     }
 
-    /// API: groups/member::invite_needs_send
+    /// API: groups/member::invite_not_sent
     ///
-    /// Returns whether the user needs an invite sent to them.  Returns true if so (whether or
-    /// not that invitation has been sent).
+    /// Returns true if there has never been an attempt to send an invitation to the user.  Returns
+    /// false otherwise.
     ///
     /// Inputs: none
     ///
     /// Outputs:
     /// - `bool` -- true if the user needs an invitation to be sent, false otherwise.
-    bool invite_needs_send() const { return invite_status == INVITE_NOT_SENT; }
+    bool invite_not_sent() const { return invite_status == INVITE_NOT_SENT; }
 
     /// API: groups/member::invite_pending
     ///
@@ -197,25 +194,25 @@ struct member {
         promotion_status = INVITE_FAILED;
     }
 
-    /// API: groups/member::accept_promotion
+    /// API: groups/member::set_promotion_accepted
     ///
     /// This marks the user as having accepted a promotion to admin in the group.
-    void accept_promotion() {
+    void set_promotion_accepted() {
         admin = true;
         invite_status = 0;
         promotion_status = 0;
     }
 
-    /// API: groups/member::promotion_needs_send
+    /// API: groups/member::promotion_not_sent
     ///
-    /// Returns whether the user needs a promotion to admin status to be sent.
-    /// Returns true if so (whether or not that promotion has failed).
+    /// Returns true if the user is an admin but there has never been an attempt to send a promotion
+    /// to admin status sent to them. Returns false otherwise.
     ///
     /// Inputs: None
     ///
     /// Outputs:
     /// - `bool` -- true if the user needs a promotion to be sent, false otherwise.
-    bool promotion_needs_send() const { return promotion_status == INVITE_NOT_SENT; }
+    bool promotion_not_sent() const { return admin && promotion_status == INVITE_NOT_SENT; }
 
     /// API: groups/member::promotion_pending
     ///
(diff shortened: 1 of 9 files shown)


src/config/groups/keys.cpp
```diff
@@ -1432,6 +1432,14 @@ LIBSESSION_C_API int groups_keys_init(
     return SESSION_ERR_NONE;
 }
 
+LIBSESSION_C_API void groups_keys_free(config_group_keys* conf) {
+    delete conf;
```
simophin, line 1437: Is it enough to just delete the `conf`? `conf->internals` is a raw pointer it seems.

include/session/config/groups/members.hpp
```diff
@@ -127,6 +129,17 @@ struct member {
...
+    /// API: groups/member::invite_needs_send
+    ///
+    /// Returns whether the user needs an invite sent to them.  Returns true if so (whether or
+    /// not that invitation has been sent).
+    ///
+    /// Inputs: none
+    ///
+    /// Outputs:
+    /// - `bool` -- true if the user needs an invitation to be sent, false otherwise.
+    bool invite_needs_send() const { return invite_status == INVITE_NOT_SENT; }
```
Bilb: should this maybe be tweak to make it clear that it isn't returning true if the invite was already tried to be sent, but failed?

include/session/config/groups/members.hpp
```diff
@@ -155,20 +168,55 @@ struct member {
...
+    /// API: groups/member::promotion_needs_send
+    ///
+    /// Returns whether the user needs a promotion to admin status to be sent.
+    /// Returns true if so (whether or not that promotion has failed).
+    ///
+    /// Inputs: None
+    ///
+    /// Outputs:
+    /// - `bool` -- true if the user needs a promotion to be sent, false otherwise.
+    bool promotion_needs_send() const { return promotion_status == INVITE_NOT_SENT; }
```
Bilb: same comment, but this one might be a bit off with the right meaning too as this method wouldn't return true of `promotion_status == INVITE_FAILED` which is what `set_promotion_failed` does

src/config/groups/keys.cpp
mpretty-cyro: I just replicated the existing `config_free` function 😬 
```
LIBSESSION_EXPORT void config_free(config_object* conf) {
    delete conf;
}
```

include/session/config/groups/members.hpp
mpretty-cyro: Yea probably a good idea - I'll rename the functions to just be `invite_not_sent` and `promotion_not_sent` to avoid the ambiguity for now, I think down the track it'd be nicer to expose it as an enum instead of having individual functions but I wanted to avoid changing the behaviour too much at this stage

include/session/config/groups/members.hpp
mpretty-cyro: Yea, I think this one is actually a bit more misleading than the other one...

Actually, looking at the logic I've changed the behaviour of the `admin` flag from "accepted admin" to be "someone that has been promoted to admin" - I'll update the functions accordingly (I think this behaviour makes more sense as we should consider a member an admin as soon as the promotion process has started, rather than only if they've confirmed receipt of the admin keys)

src/config/groups/keys.cpp
simophin: Hmm I think you do need to delete the internals here because
1. `config_object`/`config_groupb_keys` is a plain struct (it doesn't have a destructor)
2. `internal` is created using `new` and put into a raw pointer, that means it doesn't have a smart pointer to do the clean up.

Unless I miss something I think this one will leak memory (so does the config one)

jagerman: They are all copied from an original incomplete C destruction that does indeed leak.  I've pushed a commit here to fix it (both here, and two other places).  There's another one (for the network object) in the libquic API that I'll push there.

include/session/onionreq/builder.h
```diff
@@ -25,13 +25,21 @@ typedef struct onion_request_builder_object {
 ///
 /// Constructs an onion request builder and sets a pointer to it in `builder`.
 ///
-/// When done with the object the `builder` must be destroyed by either passing the pointer to
-/// onion_request_builder_free() or onion_request_builder_build().
+/// When done with the object the `builder` must be destroyed by passing the pointer to
+/// onion_request_builder_free().
```
jagerman, line 29: This comment was wrong (there was no builder_free function, the `_build()` function doesn't free anything).  And while _build *could* free, it seems a little weird to have it do so, because then you can't make one request builder and reuse it multiple times with different payloads.

The reason I comment though: if session-ios is calling this, it is worth double-checking that it isn't actually expecting `_build` to free the instance (and also adding a call to `_free` at the appropriate place).

include/session/config/groups/members.hpp
```diff
@@ -79,14 +81,11 @@ struct member {
     ///
     /// Member variable
     ///
-    /// Flag that is set to indicate to the group that this member is an admin.
+    /// Flag that is set to indicate to the group that this member has been promoted to an admin.
```
jagerman: This wording change seems slightly off, since the group creator never gets promoted to admin, and so this seems to be implying that it should be *false* for the creator because "has been promoted to" will be false.

jagerman: Perhaps "is an admin or has an in-progress promotion to admin"

src/onionreq/builder.cpp
```diff
@@ -252,7 +256,7 @@ LIBSESSION_C_API bool onion_request_builder_build(
     assert(builder && payload_in);
 
     try {
-        auto unboxed_builder = unbox(builder);
+        auto& unboxed_builder = unbox(builder);
```
jagerman, line 259: (This was unintentionally making a copy of the C++ builder object)

jagerman (approved):
LGTM (aside from the comments left).

include/session/config/groups/members.hpp
mpretty-cyro: Ah yea, good catch - will update

include/session/onionreq/builder.h
mpretty-cyro: There was a point where I'd updated session-ios to use these but before it got merged we ended up plugging `libQuic` in so now all the onion request building happens internally on the C++ side of things - so probably good to fix but luckily it won't affect anything currently 🎉

src/config/groups/keys.cpp
mpretty-cyro: Thanks for pushing up a fix for these, and good catch @simophin 🙏

PR #100, opened 2024-08-19
Updated decrypt_ons_response to support decrypting legacy ONS values

mpretty-cyro:
As it turns out we do actually need to support them 😅


mpretty-cyro:
This has been moved across to the new repo

